Issue #3, opened 2022-07-21
Cannot resign if it's not your turn


ashelkovnykov:
Self-explanatory. This will likely involve changing the `chess.hoon` logic, as (if I recall correctly) the current implementation treats resigning as a move, and you cannot perform moves on your opponent's turn.

Issue #37, opened 2022-09-08
Add panel to see previous moves, view previous board states


ashelkovnykov:
Add a panel (I'm thinking to the left-hand side of the screen) which contains the list of moves made in the game so far. The user can click on an old move to see the board state as it was at that time. I'm referring to something like this:

![example](https://user-images.githubusercontent.com/4383244/189220228-94cd49fc-2076-4cd1-856e-b4031516ee2b.png)

The panel on the right-hand side of this screenshot from lichess (obviously, the timer & ELO rating needn't be present [yet]; however, moving the "offer draw" and "resign" buttons to this panel would be a good idea).

A lot of synergy between this feature request and the #20 feature request. Mostly frontend work, some very small backend work possible.

ashelkovnykov:
Issue submitted on behalf of `~soldel-ridlun`

rovmug-ticfyn:
Is the idea with this feature that a player could request to undo back to any particular board state? I've thought about #20 as merely undoing one move at a time. Would this be a bulk undo, or just a way to browse previous states?

ashelkovnykov:
@rovmug-ticfyn Your interpretation is correct: #20 is about sending requests to the opponent to undo one move at a time. This issue is about being able to see the move history for both players while playing, and to be able to view the state of the board after that move (just to view it; if a player is viewing a past board state, there should be no way to interact with it).

bonbud-macryg:
In the frontend sketch, this left-hand "Game Panel" actually does allow undoing an arbitrary no. of moves. When the player is viewing a previous state, an "Undo" button appears below the panel. If clicked, it sends a request to the opponent to undo to that move.

Implementing this right away (so, no stopgap solution where we decouple this from #20) would require amending #20 to undo *to a specific move*, maybe expressed as a `@ud` of the move number (e.g. 20) which we could use to replace the `list` of `chess-move`s in `chess-game` with its first 20 items.

(cc: @rovmug-ticfyn)

bonbud-macryg:
We'll have to be mindful of frontend-backend syncing in this issue.

This panel could update with every new move from our opponent, but what if we make a move and it errors out on our ship? Ideally our move is only added to the panel/chessboard after it's been successfully computed on the backend, but how long does that take? It would cause a UX issue; anything that has to happen *fast* should happen on the frontend.

[...]

Issue #6, opened 2022-07-21
Allow users to browse completed games


ashelkovnykov:
`%chess` stores completed games, but only exposes them through `peek`s. I'm not actually sure if the Urbit http-api allows `peek`s, or what the secure way to implement this feature is (since you don't want foreign ships being able to read your past games at will). Definitely requires frontend work, may require backend work. Will also fix issue #2 in the process.

brbenji:
@ashelkovnykov For this issue. Do we want players to be able to simply have a list of archived games to see the metadata for? Or do we want to allow players to pull up archived games and then step through every move of that game on the practice board?

bonbud-macryg:
It should be trivial to scroll/"step" through every move of a game once we implement #37, which does the same for the game you're currently playing.

ashelkovnykov:
@brbenji The latter - as @bonbud-macryg pointed out in the comments for #2, there's no hard dependency between #6 & #2 and #37, but it would be easier to complete them after #37 is merged.

ashelkovnykov:
Thought occurred to me: we're probably going to want to handle this with scries and not subscriptions. As the number of games a user has played increases, fetching literally all of them on startup up, including every move and FEN position, will be untenable.

bonbud-macryg:
Yeah, I mentioned this issue in the "#37, \#2, and \#6 Plan" post in Design Notes but didn't suggest any particular solution. We might want to do a little of both.

Maybe the frontend subscribes to a wire that'll inform it of new games added to `completed-games`; when it first subscribes, `on-watch` just returns the result of scrying the most recent (let's say) 25 items in `completed-games`. When the player scrolls to the bottom of that list, it scries the next 25.

Question there is how best to tell the backend which range of games we want. The only thing that occurs to me right now is an arm which takes a range (or a starting index: how many completed games are currently loaded on the frontend) and scries that range in `completed-games`.

[A section of App School II](https://developers.urbit.org/guides/core/app-school-full-stack/7-app-logic#getting-entries) covers a similar problem in some detail, on loading a certain number of journal entries at one time. Linked page is the actual code but some of the thinking behind it might be in earlier pages.

Issue #5, opened 2022-07-21
Cannot claim 50-move rule draw


ashelkovnykov:
Self-explanatory. The logic is there for it in `chess.hoon`, but the frontend does not offer players a way to claim the draw.

PR #40, opened 2022-09-09
Implement threefold draw on the backend

### LINKED ISSUE #4: Cannot claim threefold repetition draw
Self-explanatory. Will require modifying `chess.hoon`.

rovmug-ticfyn:
This partially resolves issue #4, but I would like a review before moving forward with the frontend

### DIFF
--- a/src/frontend/ts/helpers/urbitChess.ts
+++ b/src/frontend/ts/helpers/urbitChess.ts
@@ -1,5 +1,5 @@
 import Urbit from '@urbit/http-api'
-import { Side, CastleSide, PromotionRole, Result, Action, MoveActionAction, GameID, Rank, File, Ship, ChessAction, ChessChallengeAction, ChessAcceptAction, ChessDeclineAction, ChessGameAction, OfferDrawAction, AcceptDrawAction, DeclineDrawAction, MoveAction, MoveMoveAction, CastleMoveAction, EndMoveAction } from '../types/urbitChess'
+import { Side, CastleSide, PromotionRole, Result, Action, MoveActionAction, GameID, Rank, File, Ship, ChessAction, ChessChallengeAction, ChessAcceptAction, ChessDeclineAction, ChessGameAction, OfferDrawAction, AcceptDrawAction, DeclineDrawAction, MoveAction, MoveMoveAction, CastleMoveAction, EndMoveAction, ChangeSpecialDrawPreferenceAction, ClaimSpecialDrawAction } from '../types/urbitChess'
 
 function emptyFunction (): void {}
 
@@ -40,6 +40,16 @@ export function declineGame (who: Ship) {
   const action: ChessDeclineAction = {
     'chess-action': Action.DeclineGame,
     'who': who
+  }
+
+  return action
+}
+
+export function changeSpecialDrawPreference (gameId: GameID, setting: boolean): ChangeSpecialDrawPreferenceAction {
+  const action: ChangeSpecialDrawPreferenceAction = {
+    'chess-action': Action.ChangeSpecialDrawPreference,
+    'game-id': gameId,
+    'setting': setting
   }
 
   return action
@@ -113,4 +123,13 @@ export function declineDraw (gameId: GameID): DeclineDrawAction {
   }
 
   return move
+}
+
+export function claimSpecialDraw (gameId: GameID): ClaimSpecialDrawAction {
+  const action: ClaimSpecialDrawAction = {
+    'chess-action': Action.ClaimSpecialDraw,
+    'game-id': gameId
+  }
+
+  return action
 }
--- a/src/frontend/ts/state/chessStore.ts
+++ b/src/frontend/ts/state/chessStore.ts
@@ -1,7 +1,7 @@
 import create from 'zustand'
 import Urbit from '@urbit/http-api'
 import { CHESS } from '../constants/chess'
-import { Update, Ship, GameID, GameInfo, ActiveGameInfo, Challenge, ChessUpdate, ChallengeUpdate, PositionUpdate, ResultUpdate, DrawOfferUpdate, DrawDeclinedUpdate } from '../types/urbitChess'
+import { Update, Ship, GameID, GameInfo, ActiveGameInfo, Challenge, ChessUpdate, ChallengeUpdate, PositionUpdate, ResultUpdate, DrawOfferUpdate, DrawDeclinedUpdate, SpecialDrawPreferenceUpdate } from '../types/urbitChess'
 import ChessState from './chessState'
 
 const useChessStore = create<ChessState>((set, get) => ({
@@ -20,6 +20,8 @@ const useChessStore = create<ChessState>((set, get) => ({
       position: CHESS.defaultFEN,
       gotDrawOffer: false,
       sentDrawOffer: false,
+      drawClaimAvailable: false,
+      autoClaimSpecialDraws: false,
       info: data
     }
 
@@ -46,12 +48,15 @@ const useChessStore = create<ChessState>((set, get) => ({
       case Update.Position: {
         const positionData = data as PositionUpdate
         const gameID = positionData.gameID
+        const specialDrawAvailable = positionData.specialDrawAvailable
 
         const currentGame = get().activeGames.get(gameID)
         const updatedGame: ActiveGameInfo = {
           position: positionData.position,
           gotDrawOffer: currentGame.gotDrawOffer,
           sentDrawOffer: currentGame.sentDrawOffer,
+          drawClaimAvailable: specialDrawAvailable,
+          autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
           info: currentGame.info
         }
 
@@ -87,6 +92,8 @@ const useChessStore = create<ChessState>((set, get) => ({
           position: currentGame.position,
           gotDrawOffer: true,
           sentDrawOffer: currentGame.sentDrawOffer,
+          drawClaimAvailable: currentGame.drawClaimAvailable,
+          autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
           info: currentGame.info
         }
 
@@ -105,6 +112,8 @@ const useChessStore = create<ChessState>((set, get) => ({
           position: currentGame.position,
           gotDrawOffer: currentGame.gotDrawOffer,
           sentDrawOffer: false,
+          drawClaimAvailable: currentGame.drawClaimAvailable,
+          autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
           info: currentGame.info
         }
 
@@ -112,6 +121,27 @@ const useChessStore = create<ChessState>((set, get) => ({
         updateDisplayGame(updatedGame)
 
         console.log('RECEIVED DRAW DECLINE UPDATE')
+        break
+      }
+      case Update.SpecialDrawPreference: {
+        const preferenceData = data as SpecialDrawPreferenceUpdate
+        const gameID = preferenceData.gameID
+        const setting = preferenceData.setting
+
+        const currentGame = get().activeGames.get(gameID)
+        const updatedGame: ActiveGameInfo = {
+          position: currentGame.position,
+          gotDrawOffer: currentGame.gotDrawOffer,
+          sentDrawOffer: currentGame.sentDrawOffer,
+          drawClaimAvailable: currentGame.drawClaimAvailable,
+          autoClaimSpecialDraws: setting,
+          info: currentGame.info
+        }
+
+        set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
+        updateDisplayGame(updatedGame)
+
+        console.log('RECEIVED SPECIAL DRAW PREFERENCE UPDATE')
         break
       }
       default: {
--- a/src/frontend/ts/types/urbitChess.ts
+++ b/src/frontend/ts/types/urbitChess.ts
@@ -32,7 +32,8 @@ export enum Update {
   Position = 'position',
   Result = 'result',
   DrawOffer = 'draw-offer',
-  DrawDeclined = 'draw-declined'
+  DrawDeclined = 'draw-declined',
+  SpecialDrawPreference = 'special-draw-preference'
 }
 
 export enum Action {
@@ -42,7 +43,9 @@ export enum Action {
   OfferDraw = 'offer-draw',
   AcceptDraw = 'accept-draw',
   DeclineDraw = 'decline-draw',
-  Move = 'move'
+  Move = 'move',
+  ChangeSpecialDrawPreference = 'change-special-draw-preference',
+  ClaimSpecialDraw = 'claim-special-draw'
 }
 
 // Yes, I know how retarded 'MoveActionAction.Move' looks - thank Ray for the types in sur/chess.hoon
@@ -79,6 +82,8 @@ export type ActiveGameInfo = {
   position: FENPosition,
   gotDrawOffer: boolean,
   sentDrawOffer: boolean,
+  drawClaimAvailable: boolean,
+  autoClaimSpecialDraws: boolean,
   info: GameInfo
 }
 
@@ -110,6 +115,7 @@ export interface PositionUpdate extends ChessUpdate {
   chessUpdate: Update.Position
   gameID: GameID
   position: FENPosition
+  specialDrawAvailable: boolean
 }
 
 export interface ResultUpdate extends ChessUpdate {
@@ -126,6 +132,12 @@ export interface DrawOfferUpdate extends ChessUpdate {
 export interface DrawDeclinedUpdate extends ChessUpdate {
   chessUpdate: Update.DrawDeclined
   gameID: GameID
+}
+
+export interface SpecialDrawPreferenceUpdate extends ChessUpdate {
+  chessUpdate: Update.SpecialDrawPreference
+  gameID: GameID
+  setting: boolean
 }
 
 // Actions
@@ -165,6 +177,15 @@ export interface AcceptDrawAction extends ChessGameAction {
 
 export interface DeclineDrawAction extends ChessGameAction {
   'chess-action': Action.DeclineDraw
+}
+
+export interface ChangeSpecialDrawPreferenceAction extends ChessGameAction {
+  'chess-action': Action.ChangeSpecialDrawPreference
+  'setting': boolean
+}
+
+export interface ClaimSpecialDrawAction extends ChessGameAction {
+  'chess-action': Action.ClaimSpecialDraw
 }
 
 // Moves
--- a/src/frontend/tsx/Chessboard.tsx
+++ b/src/frontend/tsx/Chessboard.tsx
@@ -10,7 +10,7 @@ import { CHESSGROUND } from '../ts/constants/chessground'
 import { URBIT_CHESS } from '../ts/constants/urbitChess'
 import { getChessDests, isChessPromotion } from '../ts/helpers/chess'
 import { getCgColor } from '../ts/helpers/chessground'
-import { pokeAction, move, castle, acceptDraw, declineDraw } from '../ts/helpers/urbitChess'
+import { pokeAction, move, castle, acceptDraw, declineDraw, claimSpecialDraw } from '../ts/helpers/urbitChess'
 import useChessStore from '../ts/state/chessStore'
 import { PromotionMove } from '../ts/types/chessground'
 import { Side, CastleSide, PromotionRole, Rank, File, GameID, ActiveGameInfo } from '../ts/types/urbitChess'
@@ -273,6 +273,11 @@ export function Chessboard () {
   const declineDrawOnClick = async () => {
     const gameID = displayGame.info.gameID
     await pokeAction(urbit, declineDraw(gameID), null, () => { declinedDraw(gameID) })
+  }
+
+  const acceptSpecialDrawOnClick = async () => {
+    const gameID = displayGame.info.gameID
+    await pokeAction(urbit, claimSpecialDraw(gameID))
   }
 
   //
--- a/src/frontend/tsx/Games.tsx
+++ b/src/frontend/tsx/Games.tsx
@@ -4,7 +4,7 @@ import { Side, GameID, GameInfo, ActiveGameInfo } from '../ts/types/urbitChess'
 import useChessStore from '../ts/state/chessStore'
 
 export function Games () {
-  const { urbit, displayGame, activeGames, setDisplayGame, offeredDraw, practiceBoard, setPracticeBoard } = useChessStore()
+  const { urbit, displayGame, activeGames, setDisplayGame } = useChessStore()
   const hasGame: boolean = (displayGame !== null)
 
   const extractDate = (gameID: GameID) => {
--- a/src/frontend/tsx/Settings.tsx
+++ b/src/frontend/tsx/Settings.tsx
@@ -1,6 +1,17 @@
 import React from 'react'
+import { pokeAction, changeSpecialDrawPreference } from '../ts/helpers/urbitChess'
+import useChessStore from '../ts/state/chessStore'
 
 export function Settings () {
+  const { urbit, displayGame, activeGames } = useChessStore()
+  const hasGame: boolean = (displayGame !== null)
+
+  const handleCheckboxChange = async () => {
+    const newAutoClaimPreference = !displayGame.autoClaimSpecialDraws
+    const gameID = displayGame.info.gameID
+    await pokeAction(urbit, changeSpecialDrawPreference(gameID, newAutoClaimPreference))
+  }
+
   return (
     <div className='settings-container'>
       <div id="visuals-settings" className="control-panel-container col">
@@ -8,6 +19,17 @@ export function Settings () {
       </div>
       <div id="gameplay-settings" className="control-panel-container col">
         <h4 className="control-panel-header">Gameplay</h4>
+        {hasGame ? (
+          <label>
+            <input
+              type="checkbox"
+              checked={displayGame.autoClaimSpecialDraws}
+              onChange={handleCheckboxChange}
+            />
+            Auto-Claim Special Draws
+          </label>
+        ) : (null)
+        }
       </div>
       <div id="data-settings" className="control-panel-container col">
         <h4 className="control-panel-header">Data</h4>
--- a/src/urbit/app/chess.hoon
+++ b/src/urbit/app/chess.hoon
@@ -1,23 +1,27 @@
 ::  chess: fully decentralized, peer-to-peer chess app for urbit
 ::
 ::  import libraries and expose namespace
-/+  chess, dbug, default-agent
-=,  chess
+/-  *historic
+/+  *chess, dbug, default-agent
 ::
 ::  define state structures
 |%
 +$  versioned-state
   $%  state-0
+      state-1
   ==
 +$  active-game-state
   $:  game=chess-game
       position=chess-position
+      fen-repetition=(map @t @ud)
+      special-draw-available=?
       ready=?
       sent-draw-offer=?
       got-draw-offer=?
+      auto-claim-special-draws=?
   ==
-+$  state-0
-  $:  %0
++$  state-1
+  $:  %1
       games=(map @dau active-game-state)
       archive=(map @dau chess-game)
       challenges-sent=(map ship chess-challenge)
@@ -27,7 +31,7 @@
 +$  card  card:agent:gall
 --
 %-  agent:dbug
-=|  state-0
+=|  state-1
 =*  state  -
 ^-  agent:gall
 =<
@@ -62,7 +66,8 @@
   ^-  (quip card _this)
   =/  old-state  !<(versioned-state old-state-vase)
   ?-  -.old-state
-    %0  [~ this(state old-state)]
+    %1  [~ this(state old-state)]
+    %0  [~ this(state *state-1)]
   ==
 ++  on-poke
   |=  [=mark =vase]
@@ -156,7 +161,7 @@
                                (~(del by challenges-sent) our.bowl)
                              challenges-sent
             ::  put our new game into the map of games
-            games  (~(put by games) game-id [new-game *chess-position | | |])
+            games  (~(put by games) game-id [new-game *chess-position *(map @t @ud) | | | | |])
           ==
         %decline-game
           =/  challenge  (~(get by challenges-received) who.action)
@@ -174,6 +179,24 @@
           %=  this
             ::  remove our challenger from challenges-received
             challenges-received  (~(del by challenges-received) who.action)
+          ==
+        %change-special-draw-preference
+          =/  game  (~(get by games) game-id.action)
+          ?~  game
+            :_  this
+            =/  err
+              "no active game with id {<game-id.action>}"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          :-  :~  :*  %give
+                      %fact
+                      ~[/game/(scot %da game-id.action)/updates]
+                      %chess-update
+                      !>([%special-draw-preference game-id.action setting.action])
+                  ==
+              ==
+          %=  this
+            games  (~(put by games) game-id.action u.game(auto-claim-special-draws setting.action))
           ==
         %offer-draw
           =/  game  (~(get by games) game-id.action)
@@ -257,6 +280,55 @@
             ::  record that draw offer is gone
             games  (~(put by games) game-id.action u.game(got-draw-offer |))
           ==
+        %claim-special-draw
+          =/  game-state  (~(get by games) game-id.action)
+          ?~  game-state
+            :_  this
+            =/  err
+              "no active game with id {<game-id.action>}"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          =/  ship-to-move
+            ?-  player-to-move.position.u.game-state
+              %white
+                white.game.u.game-state
+              %black
+                black.game.u.game-state
+            ==
+          ::  check whether it's our turn
+          ?>  ?=([%ship @p] ship-to-move)
+          ?.  (team:title +.ship-to-move src.bowl)
+            :_  this
+            =/  err
+              "cannot claim special draw on opponent's turn"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          ::  check if a special draw claim is available
+          ?.  special-draw-available.u.game-state
+            :_  this
+            =/  err
+              "no special draw available for game {<game-id.action>}"
+              :~  [%give %poke-ack `~[leaf+err]]
+              ==
+          :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
+                      %chess-game-result  !>([game-id.action %'½–½' ~])
+                  ==
+                  :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
+                      %chess-update
+                      !>([%result game-id.action %'½–½'])
+                  ==
+                  :*  %give  %kick  :~  /game/(scot %da game-id.action)/updates
+                                        /game/(scot %da game-id.action)/moves
+                                    ==
+                      ~
+                  ==
+              ==
+          =/  updated-game  game.u.game-state
+          =.  result.updated-game  `(unit chess-result)``%'½–½'
+          %=  this
+            games    (~(del by games) game-id.action)
+            archive  (~(put by archive) game-id.action updated-game)
+          ==
         %move
           =/  game-state  (~(get by games) game-id.action)
           ::  check for valid game
@@ -290,7 +362,7 @@
             :~  [%give %poke-ack `~[leaf+err]]
             ==
           =/  move-result
-            (try-move game.u.game-state position.u.game-state move.action)
+            (try-move u.game-state move.action)
           ::  check the move is legal
           ?~  new.move-result
             :_  this
@@ -301,8 +373,11 @@
             :~  [%give %poke-ack `~[leaf+err]]
             ==
           =,  u.new.move-result
-          :-  :_  cards.move-result
+          :-  ?:  &(auto-claim-special-draws special-draw-available)
+                ::  don't send extra move card if auto-claiming special draw
+                cards.move-result
               ::  send move to opponent
+              :_  cards.move-result
               :*  %give
                   %fact
                   ~[/game/(scot %da game-id.action)/moves]
@@ -319,7 +394,7 @@
           ::  otherwise, update position
           %=  this
             games  %+  ~(put by games)  game-id.action
-                   [game position |2.u.game-state]
+                   u.new.move-result
           ==
       ==
     ::
@@ -416,7 +491,7 @@
                                  (~(del by challenges-sent) our.bowl)
                                challenges-sent
               rng-state  (~(del by rng-state) src.bowl)
-              games  (~(put by games) game-id [new-game *chess-position | | |])
+              games  (~(put by games) game-id [new-game *chess-position *(map @t @ud) | | | | |])
             ==
           ::  we're the challenger
           ?>  ?=(^ her-hash.u.commitment)
@@ -451,11 +526,11 @@
         ==
       =/  fen  (position-to-fen u.new-position)
       :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
-                  %chess-update  !>([%position game-id.action fen])
+                  %chess-update  !>([%position game-id.action fen |])
               ==
           ==
       %=  this
-        games  (~(put by games) game-id.action [game.action u.new-position & | |])
+        games  (~(put by games) game-id.action [game.action u.new-position *(map @t @ud) | & | | |])
       ==
     ::
     ::  directly inject game subscriptions (for debugging)
@@ -574,7 +649,7 @@
         %=  this
           challenges-sent  (~(del by challenges-sent) src.bowl)
           rng-state  (~(del by rng-state) src.bowl)
-          games  (~(put by games) u.game-id [new-game *chess-position & | |])
+          games  (~(put by games) u.game-id [new-game *chess-position *(map @t @ud) | & | | |])
         ==
       ::  assign white and black to players if challenger chose
       =+  ^=  [white-player black-player]
@@ -607,7 +682,7 @@
           ==
       %=  this
         challenges-sent  (~(del by challenges-sent) src.bowl)
-        games  (~(put by games) u.game-id [new-game *chess-position & | |])
+        games  (~(put by games) u.game-id [new-game *chess-position *(map @t @ud) | & | | |])
       ==
     [%game @ta %updates ~]
       =/  game-id  `(unit @dau)`(slaw %da i.t.path)
@@ -627,7 +702,7 @@
       =/  fen  (position-to-fen position.game-state)
       =/  cards  ^-  (list card)
         :~  :*  %give  %fact  ~[/game/(scot %da u.game-id)/updates]
-                %chess-update  !>([%position u.game-id fen])
+                %chess-update  !>([%position u.game-id fen special-draw-available.game-state])
             ==
         ==
       =?  cards  got-draw-offer.game-state
@@ -750,7 +825,7 @@
                 [~ this]  :: nice try, cheater
               =/  move  !<(chess-move q.cage.sign)
               =/  move-result
-                (try-move game.u.game-state position.u.game-state move)
+                (try-move u.game-state move)
               ::  illegal move
               ?~  new.move-result
                 [cards.move-result this]  ::  nice try, cheater
@@ -765,7 +840,38 @@
               ::  add new games to our list
               %=  this
                 games  %+  ~(put by games)  u.game-id
-                       [game position |2.u.game-state]
+                       u.new.move-result
+              ==
+            %chess-game-result
+              =/  result  !<(chess-game-result q.cage.sign)
+              =/  result-game-state
+                ?~  move.result
+                  u.game-state
+                =/  move-result  (try-move u.game-state (need move.result))
+                ::  technically allows opponent to claim special draw with invalid move,
+                ::  but only when a special draw is already available
+                ::  so it doesn't break the game's correctness
+                ?~  new.move-result
+                  u.game-state
+                u.new.move-result
+              =,  result-game-state
+              ?.  special-draw-available
+                [~ this]  ::  nice try, cheater
+              :_  %=  this
+                    games    (~(del by games) u.game-id)
+                    archive  (~(put by archive) u.game-id game.result-game-state(result `result.result))
+                  ==
+              :~  :*  %give
+                      %fact
+                      ~[/game/(scot %da u.game-id)/updates]
+                      %chess-update
+                      !>([%result u.game-id result.result])
+                  ==
+                  :*  %give  %kick  :~  /game/(scot %da u.game-id)/updates
+                                        /game/(scot %da u.game-id)/moves
+                                    ==
+                      ~
+                  ==
               ==
           ==
       ==
@@ -778,34 +884,36 @@
 ::  helper core for moves
 ::  test if a given move is legal
 ++  try-move
-  |=  [game=chess-game position=chess-position move=chess-move]
-  ^-  [new=(unit [game=chess-game position=chess-position]) cards=(list card)]
-  ?.  ?=(~ result.game)
+  |=  [game-state=active-game-state move=chess-move]
+  ^-  [new=(unit active-game-state) cards=(list card)]
+  ?.  ?=(~ result.game.game-state)
     [~ ~]
   =/  new-position
-    (~(apply-move with-position position) move)
+    (~(apply-move with-position position.game-state) move)
   ?~  new-position
     [~ ~]
-  =/  updated-game  `chess-game`game
+  =/  updated-game  `chess-game`game.game-state
   =.  moves.updated-game  (snoc moves.updated-game move)
-  =/  fen  (position-to-fen u.new-position)
-  =/  position-update-card
-    :*  %give  %fact  ~[/game/(scot %da game-id.game)/updates]
-        %chess-update  !>([%position game-id.game fen])
-    ==
+  =/  new-fen-repetition  (increment-repetition fen-repetition.game-state u.new-position)
   =/  in-checkmate  ~(in-checkmate with-position u.new-position)
   =/  in-stalemate  ?:  in-checkmate
                       |
                     ~(in-stalemate with-position u.new-position)
-  ::  check if game ends by checkmate or stalemate
-  ?:  ?|  ?=(%end -.move)
-          in-checkmate
+  =/  special-draw-available  (check-threefold new-fen-repetition u.new-position)
+  =/  special-draw-claim  &(special-draw-available auto-claim-special-draws.game-state)
+  =/  position-update-card
+  :*  %give  %fact  ~[/game/(scot %da game-id.game.game-state)/updates]
+      %chess-update  !>([%position game-id.game.game-state (position-to-fen u.new-position) special-draw-available])
+  ==
+  ::  check if game ends by checkmate, stalemate, or special draw
+  ?:  ?|  in-checkmate
           in-stalemate
+          special-draw-claim
       ==
       ::  update result with score
       =.  result.updated-game
-        ?:  ?=(%end -.move)  `+.move
         ?:  in-stalemate  `%'½–½'
+        ?:  special-draw-claim  `%'½–½'
         ?:  in-checkmate
           ?-  player-to-move.u.new-position
             %white  `%'0-1'
@@ -813,21 +921,53 @@
           ==
         !!
       ::  give a card of the game result to opponent ship
-      :-  `[updated-game u.new-position]
+      :-  `[updated-game u.new-position new-fen-repetition special-draw-available |4.game-state]
+      ?.  special-draw-claim
+        :~  position-update-card
+            :*  %give  %fact  ~[/game/(scot %da game-id.game.game-state)/updates]
+                %chess-update
+                !>([%result game-id.game.game-state (need result.updated-game)])
+            ==
+            ::  kick subscriber from game
+            :*  %give  %kick  :~  /game/(scot %da game-id.game.game-state)/updates
+                                  /game/(scot %da game-id.game.game-state)/moves
+                              ==
+                ~
+            ==
+        ==
+      ::  if we're auto-claiming a special draw, send opponent our move with the result
       :~  position-update-card
-          :*  %give  %fact  ~[/game/(scot %da game-id.game)/updates]
-              %chess-update
-              !>([%result game-id.game (need result.updated-game)])
+          :*  %give  %fact  ~[/game/(scot %da game-id.game.game-state)/moves]
+              %chess-game-result
+              !>([game-id.game.game-state (need result.updated-game) `move])
           ==
-          ::  kick subscriber from game
-          :*  %give  %kick  :~  /game/(scot %da game-id.game)/updates
-                                /game/(scot %da game-id.game)/moves
+          :*  %give  %fact  ~[/game/(scot %da game-id.game.game-state)/updates]
+              %chess-update
+              !>([%result game-id.game.game-state (need result.updated-game)])
+          ==
+          :*  %give  %kick  :~  /game/(scot %da game-id.game.game-state)/updates
+                                /game/(scot %da game-id.game.game-state)/moves
                             ==
               ~
           ==
       ==
-  ::  inform opponent of new position
-  :-  `[updated-game u.new-position]
+  :-  `[updated-game u.new-position new-fen-repetition special-draw-available |4.game-state]
   :~  position-update-card
   ==
+++  increment-repetition
+  |=  [fen-repetition=(map @t @ud) position=chess-position]
+  ^-  (map @t @ud)
+  =/  fen  ~(simplified position-to-fen position)
+  =*  count  (~(get by fen-repetition) fen)
+  ?~  count
+    (~(put by fen-repetition) fen 1)
+  (~(put by fen-repetition) fen +((need count)))
+++  check-threefold
+  |=  [fen-repetition=(map @t @ud) position=chess-position]
+  ^-  ?
+  =/  fen  ~(simplified position-to-fen position)
+  =*  count  (~(get by fen-repetition) fen)
+  ?~  count
+    |
+  (gth (need count) 2)
 --
--- a/src/urbit/lib/chess.hoon
+++ b/src/urbit/lib/chess.hoon
@@ -327,7 +327,7 @@
 ::  as forsyth-edwards notation
 ++  position-to-fen
   |_  chess-position
-  ++  $
+  ++  simplified
     ^-  @t
     ;:  (cury cat 3)
       board-to-fen
@@ -342,6 +342,11 @@
       ?~  en-passant
         '-'
       (square-to-algebraic u.en-passant)
+    ==
+  ++  $
+    ^-  @t
+    ;:  (cury cat 3)
+      simplified
       ' '
       (scot %ud ply-50-move-rule)
       ' '
@@ -762,11 +767,8 @@
   ++  algebraicize
     |=  move=chess-move
     ^-  @t
-    ?>  ?|  ?=(%end -.move)
-            (legal-move move)
-        ==
+    ?>  (legal-move move)
     ?-  -.move
-      %end  +.move
       %castle
         ?-  +.move
           %queenside  'O-O-O'
@@ -840,7 +842,6 @@
     ++  $
       ^-  chess-position
       ?-  -.move
-        %end     position
         %castle
           %=  position
             board             %+  ~(castle with-board board)
@@ -1037,11 +1038,6 @@
     ^-  ?
     &(!in-check no-legal-moves)
   ::
-  ::  XX no threefold repetition yet
-  ++  draw-claimable
-    ^-  ?
-    (legal-move [%end %'½–½'])
-  ::
   ::  legal move logic
   ::
   ::  the list of legal moves is empty
@@ -1065,14 +1061,6 @@
     ++  $
       ^-  ?
       ?-  -.move
-        %end
-          ?-  +.move
-            ::  resignations ('1-0' = white victory, '0-1' = black victory)
-            %'0-1'  =(player-to-move %white)
-            %'1-0'  =(player-to-move %black)
-            ::  check if the 50 play rule is claimable
-            %'½–½'  (gte ply-50-move-rule 100)
-          ==
         %castle  castle-move
         %move    normal-move
       ==
--- a/src/urbit/mar/chess/action.hoon
+++ b/src/urbit/mar/chess/action.hoon
@@ -48,6 +48,25 @@
         %-  ot:dejs
         :~  [%game-id (se:dejs %da)]
         ==
+      %change-special-draw-preference
+        :-  %change-special-draw-preference
+        %.  jon
+        %-  ot:dejs
+        :~  [%game-id (se:dejs %da)]
+            [%setting bo:dejs]
+        ==
+      %claim-special-draw
+        :-  %claim-special-draw
+        %.  jon
+        %-  ot:dejs
+        :~  [%game-id (se:dejs %da)]
+        ==
+      %decline-special-draw
+        :-  %decline-special-draw
+        %.  jon
+        %-  ot:dejs
+        :~  [%game-id (se:dejs %da)]
+        ==
       %move
         :-  %move
         =+  ^-  [game-id=@dau head=@tas]
@@ -83,13 +102,6 @@
             %.  jon
             %-  ot:dejs
             :~  [%castle-side so:dejs]
-            ==
-          %end
-            :-  game-id
-            :-  %end
-            %.  jon
-            %-  ot:dejs
-            :~  [%result so:dejs]
             ==
         ==
     ==
--- a/src/urbit/mar/chess/update.hoon
+++ b/src/urbit/mar/chess/update.hoon
@@ -23,6 +23,7 @@
         :~  ['chessUpdate' [%s 'position']]
             ['gameID' [%s (scot %da game-id.upd)]]
             ['position' [%s position.upd]]
+            ['specialDrawAvailable' [%b special-draw-available.upd]]
         ==
       %result
         %-  pairs:enjs
@@ -39,6 +40,12 @@
         %-  pairs:enjs
         :~  ['chessUpdate' [%s 'draw-declined']]
             ['gameID' [%s (scot %da game-id.upd)]]
+        ==
+      %special-draw-preference
+        %-  pairs:enjs
+        :~  ['chessUpdate' [%s 'special-draw-preference']]
+            ['gameID' [%s (scot %da game-id.upd)]]
+            ['setting' [%b setting.upd]]
         ==
     ==
   --
--- a/src/urbit/sur/chess.hoon
+++ b/src/urbit/sur/chess.hoon
@@ -165,10 +165,8 @@
 ::  a queen- or king-side castle;
 ::  or a finishing move with a result
 +$  chess-move
-  $~  [%end %'½–½']
   $%  [%move from=chess-square to=chess-square into=(unit chess-promotion)]
       [%castle ?(%queenside %kingside)]
-      [%end chess-result]
   ==
 ::
 ::  chess-game stores metadata for a game
@@ -179,7 +177,7 @@
           event='?'
           site='Urbit Chess'
           date=*@da
-          ::  the round's default value is 
+          ::  the round's default value is
           ::  ~ if unknown, `~ if inappropriate
           round=~
           white=*chess-player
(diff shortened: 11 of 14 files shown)


ashelkovnykov:
You've accidentally merged your changes from your pending documentation PR into this one. Please separate the changes so that this PR only includes backend work for the threefold draw bug.

rovmug-ticfyn:
At this point I need help with the `on-load` arm

src/urbit/lib/chess.hoon
```diff
@@ -345,6 +345,78 @@
...
+        %kingside   'K'
+        %none       ''
+      ==
+      ?-  black-can-castle
+        %both       'kq'
+        %queenside  'q'
+        %kingside   'k'
+        %none       ''
+      ==
+  --
```
ashelkovnykov: There is a much simpler way to do this that doesn't require copying a large portion of the code:
```
++  position-to-fen
  |_  chess-position
  ++  simplified
    ^-  @t
    ;:  (cury cat 3)
      board-to-fen
      ' '
      ?-  player-to-move
        %white  'w'
        %black  'b'
      ==
      ' '
      castle-to-fen
      ' '
      ?~  en-passant
        '-'
      (square-to-algebraic u.en-passant)
    ==
  ++  $
    ^-  @t
    ;:  (cury cat 3)
      simplified
      ' '
      (scot %ud ply-50-move-rule)
      ' '
      (scot %ud move-number)
    ==
  ++  board-to-fen
    ...
```
Then you would call the above as:
```
~(simplified position-to-fen chess-position)
```

src/urbit/sur/chess.hoon
```diff
@@ -173,4 +176,16 @@
...
+::  $~  [*chess-board %white %both %both ~ 0 1]
+::  $:
+::    board=chess-board
+::    player-to-move=chess-side
+::    white-can-castle=chess-castle
+::    black-can-castle=chess-castle
+::    en-passant=(unit chess-square)
+::    ply-50-move-rule=@
+::    move-number=@
+::  ==
```
ashelkovnykov: Is this a temp arm you made for testing?

src/urbit/sur/chess.hoon
```diff
@@ -95,6 +95,8 @@
     en-passant=(unit chess-square)
     ply-50-move-rule=@
     move-number=@
+    fen-repetition=(map @t @ud)
+    threefold-draw-available=?
```
ashelkovnykov: `chess-position` is basically the `%chess` equivalent of the FEN. I can see the logic of why you might put the position repetition table into this structure: just like the move number and 50-move rule ply count, it represents something about the history of the position. However, this is not the best place to put this information. The information that would be most in keeping with the other information would be a single number representing how many times this exact position has been observed, but we can track the same information elsewhere in a more efficient manner, so I digress.

The next level up would be the `chess-game` type. However, this is an abstraction that's used both for active games and archived games. Though I could see it being useful (theoretically), we don't need to know how many times a particular position occurred in an archived game. If we really needed that information, we could recompute it from the move list. Therefore, `chess-game` is not the right location either.

That leaves us with only one place that a `map` counting the number of times each position occurred could go: `active-game-state`. This is the best place to put the `map`, as just like the `sent-draw-offer` and `got-draw-offer` fields, it would provide additional information that would help the `%chess` agent send/validate valid draw claims, and dump this information upon completion of the game.

src/urbit/mar/chess/update.hoon
```diff
@@ -40,6 +40,11 @@
         :~  ['chessUpdate' [%s 'draw-declined']]
             ['gameID' [%s (scot %da game-id.upd)]]
         ==
+      %threefold-checkbox
+        %-  pairs:enjs
+        :~  ['chessUpdate' [%s 'threefold-checkbox']]
+            ['gameID' [%s (scot %da game-id.upd)]]
+        ==
```
ashelkovnykov: A checkbox is an entirely frontend concept - there should be absolutely no mention of it in the backend at all. In my design for a solution to the threefold repetition bug, the checkbox I mentioned is just a frontend interface for communicating a particular piece of data: does the user want to automatically claim threefold repetition draws, or not? As a "yes/no" piece of information, this is a single boolean (or loobean) that the user needs to communicate to the `%chess` app.

The problem that this is solving is that the player receiving a position that's been repeated three times can choose to claim or not claim a draw from that position. However, the player who is creating a position that's been repeated three times must make the draw claim as he makes the move. Therefore, this "yes/no" decision should be provided with an accompanying move. This provides the inspiration for how we should design the interface on the backend: this is information that we should receive with a corresponding move.

What kind of move should this information be a part of? We have two options available: `chess-move` and `chess-action`. I leave it as an exercise for you to decide which one would be a better design.

src/urbit/lib/chess.hoon
```diff
@@ -865,6 +941,28 @@
...
+      %=  $
+        fen-repetition  (~(put by fen-repetition) fen 1)
+      ==
+    ?.  (gte 2 (~(got by fen-repetition) fen))
+      %=  $
+        fen-repetition  (~(put by fen-repetition) fen 2)
+      ==
+      %=  $
+        fen-repetition  (~(put by fen-repetition) fen +((~(got by fen-repetition) fen)))
+      ==
```
ashelkovnykov: This function doesn't do what it claims it does: it's not returning an updated map, it's returning the `$` arm with a modified subject (which I think actually means that it's looping forever). Aside from that, the name is misleading since it's not "checking" but updating a `map`. Further more, what it does for the app doesn't need it's own function: line `956` is really all that needs to be done to increment the count.

src/urbit/lib/chess.hoon
```diff
@@ -865,6 +941,28 @@
...
+  ++  check-threefold
+    |=  [fen-repetition=(map @t @ud) fen=@t]
+    ^-  ?
+    ?.  (~(has by fen-repetition) fen)
+      |
+    ?.  (gte 2 (~(got by fen-repetition) fen))
+      &
+      |
```
ashelkovnykov: There's a simpler way to write this, and also a bug (`gte` instead of `gth`):
```
|=  [fen-repetition=(map @t @ud) fen=@t]
  ^-  ?
  =*  count  (~(get by fen-repetition) fen)
  ?~  count
    |
  (gth 2 (need count))
```

src/urbit/app/chess.hoon
```diff
@@ -52,7 +63,8 @@
   ^-  (quip card _this)
   =/  old-state  !<(versioned-state old-state-vase)
   ?-  -.old-state
-    %0  [~ this(state old-state)]
+    %1  [~ this(state old-state)]
+    %0  [~ this(state 1+old-state)]
```
ashelkovnykov: `on-load` defines logic for how to revive an app that has been suspended. It's the counterpart to `on-save`: `on-save` defines how to convert the state of the app to a noun before going offline, and `on-load` defines how to convert a noun to application state before coming back online.

Since the process for going offline and then coming back online is also what an application does between version updates, `on-load` also defines state transition logic when the form of the application state changes between versions.

That means that when a version goes offline with state version `%0` and comes back to an application where state version `%1` is the latest state, it needs to have logic for converting its state from version `%0` to version `%`.

I recommend you re-read [chapter 4 of App School](https://developers.urbit.org/guides/core/app-school/4-lifecycle) to get a better grasp of the concept.

I'm going to quote my earlier thoughts from our conversation on Urbit (and add @bonbud-macryg & @brbenji 
 to the conversation below since it's relevant to all of you):
> Since `active-game-state` is a part of the application state (it's a convenience-type that's declared to make the `state-0` type easier to specify), any changes to it are indirect changes to `state-0`, and therefore it needs to be versioned to `state-1`. But the whole point of versioned states is for applications to be able to cleanly upgrade state using the `on-load` arm, once they're revived after an update. Therefore, it's possible that we may need to not only incrementally version `active-game-state`, but to also keep another copy of it which retains the spec it had in state-0. This also brings up an interesting question: what if we needed to do this for a type declared in `../sur/chess.hoon` (e.g. `chess-game`)?
>
> I see two designs, and which one we choose to employ depends on the specific changes: more specifically, it depends on what would make the state transition logic in `on-load` easier to implement, which is downstream of the specific changes.
>
> Design 1:
> - Modify the lesser type (e.g. `active-game-state`), do not incrementally version it
> - Bump app state version
> - In `on-load` arm, treat old instances of the lesser type as raw cells, access fields directly
  (i.e. `+6.old-game-state` instead of `position.old-game-state`)
> 
> Design 2:
> - Create a file for outdated types: `../sur/historic.hoon`
> - Copy the current version of the lesser type (e.g. `active-game-state`) to `../sur/historic.hoon`
  and append the state version at the time to its name (e.g. `active-game-state-0` for the version
  of `active-game-state` for `state-0`)
> - Modify the lesser type, keep its name as is
> - Bump app state version
> - Import `../sur/historic.hoon` into `../app/chess.hoon`
> - In `on-load` arm, refer to old instances of the lesser type using the historic types

I've been thinking about this lately, and I think that these are good designs, but that the combined changes for all of Milestone 1 will be so large and fix so many bugs that make `%chess` unpleasant to deal with currently, that we should just reset the state completely for the release. **Therefore, don't worry about implementing state transition logic: if you need to modify the state, just make `state-1` completely reset the entire app state.**

ashelkovnykov (changes requested):
Review for backend only

src/urbit/mar/chess/update.hoon
rovmug-ticfyn: I was thinking the yes/no state of the checkbox needed to be separate from a move so that the backend would automatically claim the draw for a player who receives a repeated position and has the box checked. To do that, my assumption was that `%chess` would need to keep track of the state of the checkbox.

src/urbit/app/chess.hoon
```diff
@@ -3,13 +3,26 @@
...
 +$  active-game-state
   $:  game=chess-game
       position=chess-position
+      fen-repetition=(map @t @ud)
+      threefold-draw-available=?
       ready=?
       sent-draw-offer=?
       got-draw-offer=?
+      claim-threefold-draw=?
+      opponent-claim-threefold-draw=?
```
ashelkovnykov: `threefold-draw-available`, `claim-threefold-draw`, and `opponent-claim-threefold-draw` aren't necessary.

`threefold-draw-available` is denoted by whether the current position of the board has a value of 3 or more in the `fen-repetition` `map`.

`claim-threefold-draw` is a piece of information that should either come with the move (if the player is moving into a position that is repeated 3 times) or its own move (if the player claims a draw after the opponent moves into a position that is repeated 3 times).

`opponent-claim-threefold-draw` is something that we verify as valid by checking whether the current position has been repeated 3 times when the opponent sends a draw result (i.e. add it to the current checks: did we offer a draw, is the 50-move rule active, is the threefold repetition rule active, or is the opponent trying to cheat?).

Note that if a player moves into a draw, the draw claim must be a part of the move. The current `chess-move` type needs to be extended to account for this.

src/urbit/app/chess.hoon
```diff
@@ -22,7 +35,7 @@
 +$  card  card:agent:gall
```
ashelkovnykov, line 31: Remember this comment: https://github.com/ashelkovnykov/urbit-chess/pull/40#discussion_r973034694

Since you're changing what the type `active-game-state` means, you need to either keep the old `active-game-state` as a renamed historic type, or you need to change the type of `games` in `state-0` because it no longer refers to the same `active-game-state` that it did previously.

src/urbit/app/chess.hoon
```diff
@@ -265,8 +279,24 @@
...
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
+                      %chess-update
+                      !>([%claim-threefold game-id.action])
+                  ==
+              ==    
+          %=  this
+            games  (~(put by games) game-id.action u.game(claim-threefold-draw !claim-threefold-draw.u.game))
+          ==
```
ashelkovnykov: Claiming the threefold repetition draw doesn't need to be its own `chess-action`, it's just a `chess-move` like any other. Only one thing complicates it: the opponent may need to make the move and claim the draw at the same time, which `chess-move` doesn't support right now. It needs to be extended to have the option to claim a draw as part of a regular piece move.

Generic code for handling end moves and the 50-move draw already exists in `with-position` in `../lib/chess.hoon` (called by `try-move` in `../app/chess.hoon`, itself called by both `on-poke` for our moves and `on-agent` for opponent moves).

However, this will not do for the threefold repetition draw, as `with-position` looks only at the current position in a vacuum. If we could extend the `chess-position` type to track move repetitions in a field like `ply-50-move-rule`, that would solve the issue. Unfortunately, that's not possible: tracking the 50-move rule simply entails incrementing the count when a move without a pawn or capture takes place, and resetting the count to 0 when a pawn move or capture occurs. Unlike this simple arithmetic, counting position repetitions requires comparing against past positions.

Therefore, the next best place is probably in `try-move`, which already handles some other end states (e.g. checkmate and stalemate). I see that you've already modified the `try-move` code to take the `map` of prior FEN positions. We may as well simplify the call:
```
++  try-move
  |=  [game=active-game-state move=chess-move]
  ^-  [new=(unit active-game-state) cards=(list card)]
  ...
```

src/urbit/lib/chess.hoon
```diff
@@ -274,6 +274,22 @@
...
+        %white  'w'
+        %black  'b'
+      ==
+      ' '
+      castle-to-fen
+      ' '
+      ?~  en-passant
+        '-'
+      (square-to-algebraic u.en-passant)
+    ==
```
ashelkovnykov, line 345: Reread the code example in this comment: https://github.com/ashelkovnykov/urbit-chess/pull/40#discussion_r972971479

There's a way to simplify this block and the code below it even more, so that there isn't two copies of the code which generates the first four fields of a FEN position

src/urbit/lib/chess.hoon
```diff
@@ -865,6 +881,22 @@
...
+        (~(put by fen-repetition) fen 1)
+      (~(put by fen-repetition) fen +((need count)))
+  ++  check-threefold
+    |=  [fen-repetition=(map @t @ud) position=chess-position]
+      ^-  ?
+      =/  fen  ~(simplified position-to-fen position)
+      =*  count  (~(get by fen-repetition) fen)
+      ?~  count
+        |
+      (gth 2 (need count))
```
ashelkovnykov: If we move the threefold repetition check into `try-move`, as I think we must, then these arms don't make sense here and should become siblings of `try-move`.

ashelkovnykov (changes requested):

src/urbit/mar/chess/update.hoon
ashelkovnykov: Your idea also makes sense. In that case, you would need to add a new type of poke that sets the user's preference for setting this option to true/false for the game. It works well with my design, in that the default value of this setting on the backend would be part of the not-yet-implemented user preferences (e.g. auto-claim draws, practice board state, board visual settings, etc.).

However, the name should still definitely not include "checkbox" anywhere in it, as that's a frontend interface abstraction.

src/urbit/app/chess.hoon
ashelkovnykov: Note based on our conversation here: https://github.com/ashelkovnykov/urbit-chess/pull/40#discussion_r975547508

With a slightly different name and a slightly different action, this poke could still be useful for setting an automatic preference on the backend regarding whether to automatically claim threefold repetition/50-move draws or not.

src/urbit/app/chess.hoon
ashelkovnykov: Note based on our conversation here: https://github.com/ashelkovnykov/urbit-chess/pull/40#discussion_r975547508

With a slightly different name, `claim-threefold-draw` could still be useful for storing the preference on the backend regarding whether to automatically claim threefold repetition/50-move draws or not.

src/urbit/app/chess.hoon
```diff
@@ -9,12 +9,10 @@
   $:  game=chess-game
       position=chess-position
       fen-repetition=(map @t @ud)
-      threefold-draw-available=?
       ready=?
       sent-draw-offer=?
       got-draw-offer=?
-      claim-threefold-draw=?
-      opponent-claim-threefold-draw=?
+      auto-claim-threefold-draw=?
```
ashelkovnykov: Let's change the name to `auto-claim-special-draws` (or something else, if you can think of a better name) and reuse this field for the 50-move draw.

src/urbit/app/chess.hoon
```diff
@@ -24,9 +22,16 @@
       challenges-received=(map ship chess-challenge)
       rng-state=(map ship chess-commitment)
   ==
++$  active-game-state-0
+  $:  game=chess-game
+      position=chess-position
+      ready=?
+      sent-draw-offer=?
+      got-draw-offer=?
+  ==
```
ashelkovnykov: Yeah, what you've done here with regards to `active-game-state` and `active-game-state-0` is what I was thinking. Only now I realize that it doesn't matter, since we're bunting `state-1` anyway, but I think it's still good practice. Would you kindly move `active-game-state-0` to a new `../sur/historical.hoon` file (reference for reasoning: https://github.com/ashelkovnykov/urbit-chess/pull/40#discussion_r973034694).

Building on the above: maybe we should rename `state-1` to `state` and move `state-0` to `../sur/historical.hoon` too? Might make sense to keep *everything* that used to be state but no longer is in a different file.

src/urbit/app/chess.hoon
```diff
@@ -751,19 +754,57 @@
...
+        (~(put by fen-repetition) fen 1)
+      (~(put by fen-repetition) fen +((need count)))
+  ++  check-threefold
+    |=  [fen-repetition=(map @t @ud) position=chess-position]
+      ^-  ?
+      =/  fen  ~(simplified position-to-fen position)
+      =*  count  (~(get by fen-repetition) fen)
+      ?~  count
+        |
+      (gth 2 (need count))
```
ashelkovnykov: Bug here: should be `(gth (need count) 2)`

src/urbit/app/chess.hoon
```diff
@@ -714,28 +718,27 @@
...
-        %chess-update  !>([%position game-id.game fen])
+    :*  %give  %fact  ~[/game/(scot %da game-id.game.game)/updates]
+        %chess-update  !>([%position game-id.game.game fen])
     ==
   =/  in-checkmate  ~(in-checkmate with-position u.new-position)
   =/  in-stalemate  ?:  in-checkmate
                       |
                     ~(in-stalemate with-position u.new-position)
-  =/  threefold-draw  &(=(& new-threefold-draw-available) |(=(& claim-threefold-draw) =(& opponent-claim-threefold-draw)))
+  =/  threefold-draw  &(=(& (check-threefold new-fen-repetition u.new-position)) =(& auto-claim-threefold-draw.game))
```
ashelkovnykov: You don't need to do explicit comparisons against `loobeans`/`booleans`: `=(& (some-gate some-input))` is exactly equivalent to `(some-gate some-input)`

src/urbit/app/chess.hoon
```diff
@@ -265,7 +284,18 @@
             ==
           %=  this
             games  %+  ~(put by games)  game-id.action
-                   [game position |2.u.game-state]
+                   u.new.move-result
+          ==
+        %auto-claim-threefold
```
ashelkovnykov: `%change-special-draw-preference` is probably a better action name, though very verbose. If you can think of something less wordy than that, please do - otherwise let's go with it.

Also, this is nit-picky, but I think that this code block should be after `%decline-game` and before `%offer-draw`. My reasoning is that the code is logically grouped into sections (offer challenge -> respond to challenge -> offer draw -> respond to draw -> move) and that setting game preferences feels like something that logically precedes draws & moves.

src/urbit/app/chess.hoon
```diff
@@ -265,7 +284,18 @@
...
+          =/  game  (~(get by games) game-id.action)
+          ?~  game
+            :_  this
+            =/  err
+              "no active game with id {<game-id.action>}"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          %=  this
+            games  (~(put by games) game-id.action u.game(auto-claim-threefold-draw !auto-claim-threefold-draw.u.game))
```
ashelkovnykov: I think we should make the new setting for the auto-draw preference explicit. My reasoning is that the current way you have it adds an implicit relationship between the frontend and the backend: whatever interface visualizes the state of the auto-claim preference needs to start in the same state as the `%chess` backend, or else it's possible that they enter into a reversed state (ex. backend default is 'auto-claim off', but frontend checkbox starts out checked).

src/urbit/mar/chess/action.hoon
```diff
@@ -48,6 +48,12 @@
         %-  ot:dejs
         :~  [%game-id (se:dejs %da)]
         ==
+      %auto-claim-threefold
```
ashelkovnykov: See above comment regarding name

ashelkovnykov (changes requested):
Let's consider the cases that our updated code needs to handle:
1. We make move that repeats position for third time; auto-claim disabled; opponent doesn't claim draw
2. We make move that repeats position for third time; auto-claim disabled; opponent claims draw
3. We make move that repeats position for third time; auto-claim enabled
4. Opponent makes move that repeats position for third time; opponent claims draw
5. Opponent makes move that repeats position for third time; opponent doesn't claim draw; auto-claim disabled; we don't claim draw
6. Opponent makes move that repeats position for third time; opponent doesn't claim draw; auto-claim disabled; we claim draw
7. Opponent makes move that repeats position for third time; opponent doesn't claim draw; auto-claim enabled

Logically:
- cases 1 & 5 are the same situation
- cases 2 & (6 | 7) are the same situation
- cases 3 & 4 are the same situation

Let's break down what should happen in each case:

#### 1:

- receive regular move poke
- send regular move card to `/game/<game id>/moves`
- send regular move card to `/game/<game id>/updates`
- receive regular move update

#### 2:

- receive regular move poke
- send regular move card to `/game/<game id>/moves`
- send regular move card to `/game/<game id>/updates`
- receive result move update
- verify that result is valid (position repeated 3 times)
- send kick card to `/game/<game id>/moves`
- send kick card to `/game/<game id>/updates`

#### 3:

- receive regular move poke
- **send threefold draw claim to `/game/<game id>/moves`**
- send result card to `/game/<game id>/updates`
- send kick card to `/game/<game id>/moves`
- send kick card to `/game/<game id>/updates`

#### 4:

- **receive threefold draw claim update**
- **verify draw claim**
- send kick card to `/game/<game id>/moves`
- send kick card to `/game/<game id>/updates`

#### 5:

- receive regular move update
- send regular move card to `/game/<game id>/updates`
- **send draw available card to `/game/<game id>/updates`**
- receive regular move poke
- send regular move card to `/game/<game id>/moves`
- send regular move card to `/game/<game id>/updates`

#### 6:

- receive regular move update
- send regular move card to `/game/<game id>/updates`
- **send draw available card to `/game/<game id>/updates`**
- receive draw claim poke
- send result card to `/game/<game id>/moves`
- send result card to `/game/<game id>/updates`
- send kick card to `/game/<game id>/moves`
- send kick card to `/game/<game id>/updates`

#### 7:

- receive regular move update
- send result card to `/game/<game id>/moves`
- send result card to `/game/<game id>/updates`
- send kick card to `/game/<game id>/moves`
- send kick card to `/game/<game id>/updates`

What work actually needs to be done for each case?

1. Nothing; current functionality
2. Need to add code for handling result updates on `/game/<game id>/moves`. This code should verify that the opponent is either claiming a threefold repetition draw, a 50-move rule draw, or resigning. This would involve pulling the resignation/50-move draw code out of `++legal-move:with-position`, and removing `%end` from `chess-move`.
3. Need to determine what a threefold draw claim from the person making the move looks like (see below)
4. Same as case 3 (see below)
5. Need to determine what it means for a draw claim to be available, as opposed to receiving a draw offer (see below)
6. Need to add code for sending result updates on `/game/<game id>/moves`; in addition, same needs as case 5 (see below)
7. Same as case 6

**Note 1:** The work for cases 2, 6, and 7 will probably resolve issue #3 (@bonbud-macryg) at the same time
**Note 2:** The work for cases 5 and 6 will probably make issue #5 (@bonbud-macryg) trivial to resolve
**Note 3:** These changes would also fix an existing bug that is not currently logged in the issues: sometimes a finished game is not removed for the losing player

#### What is does a threefold draw claim from the person making it look like?

The easiest thing to do would be 2 back-to-back cards on `/game/<game id>/moves`: a regular move and then a result. However, this isn't correct by the rules: according to FIDE, the draw claim needs to occur simultaneously with the move that results in a repeated position. Splitting the claim into two cards breaks this logic and results in at least two ways for player intent to be subverted by lag or malicious intent.

Therefore, I think that `chess-update` should be extended to include a `(unit chess-move)` which would be null unless a move is needed to explain a draw claim. Combined with the changes for case 2, this means that to verify a `%result` `chess-update`, the code would first apply the given `chess-move` if one is present, then check that a draw is valid by making a call to `check-threefold`.

#### What does it mean for a draw claim to be available?

We'll probably need to add a new piece of state to the frontend which tracks a draw claim being available. This new state will be set just like the existing `%draw-offer` `chess-update` sets the state for a draw offer being available. Similarly to declining a draw, if the user makes a regular move when a draw claim is available, that's the same as rejecting the draw claim, so an update will need to be sent to /game/<game id>/updates to revoke the draw claim availability.

In summary, this means adding two new `chess-updates`: `%draw-claim-available` and `%draw-claim-declined`. On the frontend, we'll need a way for the user to exercise his draw claim (e.g. a button). Let's put a pin in that for now, and focus on the backend.

There's potential here to preemptively solve this in a way that prepares us for issue #37, which also has synergy with issue #6 (which blocks issue #2). For that reason, we probably want to get @brbenji involved in this as well, down the line.

brbenji:
After reading through the discussion and code on this PR, I’m going to paint in broad strokes the larger design decisions being made. I want to do this both to give a signpost at this point and to be corrected if I’m seeing things wrong.
### state
The first big decision deals with state management. Inside the standard `state-0` we also have the helper state `active-game-state`. We wish to expand `active-game-state`, but there is no current example of an app doing this. 
In order to upgrade smoothly, the decision was to create a new structure file to hold the old states. 
This file is `../sur/historical.hoon`. 
Both `state-0` and `active-game-state-0` live in there now.

The other decision for state was to create a fresh start for `state-1` on-load, by transforming `state` into the bunt of `state-1`; wiping state clean for all the large changes expected after Milestone 1.
### settings
We hope to allow users the option of automatically claiming draws when special draw scenarios occur. `%change-special-draw-preference` is now in `chess-action` and `active-game-state` holds the preference in state.
### mar
There’s a new mark, `result.hoon`. This represents a new way of handling game endings. In order to verify a special draw scenario, we now include info on the `chess-move` that triggers a special draw scenario. When a `chess-update` comes with `%result` draw we now validate that move in the `++ try-move` arm.
### next steps
We’re hoping that the framework for implementing this threefold repetition draw, will lay down the groundwork for solving issues that involve stepping back through game moves. And much of that will require coordination with the frontend.

_Please correct me if I’m wrong._

src/urbit/app/chess.hoon
```diff
@@ -144,6 +148,17 @@
...
+        %change-special-draw-preference
+          =/  game  (~(get by games) game-id.action)
+          ?~  game
+            :_  this
+            =/  err
+              "no active game with id {<game-id.action>}"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          %=  this
+            games  (~(put by games) game-id.action u.game(auto-claim-special-draws setting.action))
```
bonbud-macryg, line 199: For my own sake: what's going on with the brackets in `u.game(auto-claim-special-draws setting.action)`? I don't think I've seen that before.

src/urbit/app/chess.hoon
```diff
@@ -213,6 +228,39 @@
...
+                                    ==
+                      ~
+                  ==
+              ==
+          =/  updated-game  game.u.game-state
+          =.  result.updated-game  `(unit chess-result)``%'½–½'
+          %=  this
+            games    (~(del by games) game-id.action)
+            archive  (~(put by archive) game-id.action updated-game)
+          ==
```
bonbud-macryg: Without having tested yet this poke looks okay, but why refer to it as a `special-draw` and not a `threefold-draw` when the poke is just checking for a threefold repetition? If I'm not missing something (I've read the previous discussion in this PR, but I might be) I think it would be clearer if it was called something like `threefold-draw` throughout.

src/urbit/app/chess.hoon
```diff
@@ -624,7 +667,7 @@
               ==
             %chess-draw-accept
               ?.  sent-draw-offer.u.game-state
-                [~ this]  ::  nice try, cheater
+                      [~ this]  ::  nice try, cheater
```
bonbud-macryg: I'm not sure why this is indented, the original version looks correct.

src/urbit/app/chess.hoon
```diff
@@ -666,7 +709,52 @@
...
                 games  %+  ~(put by games)  u.game-id
-                       [game position |2.u.game-state]
+                       u.new.move-result
+              ==
+            %chess-result
+              =/  result
+                !<([game-id=@dau result=chess-result move=(unit chess-move)] q.cage.sign)
+              ?~  move.result
+                ?.  (check-threefold fen-repetition.u.game-state position.u.game-state)
+                      [~ this]  ::  nice try, cheater
```
bonbud-macryg: I'd tab `[~ this]` over to look like:
```
?.  (check-threefold fen-repetition.u.game-state position.u.game-state)
  [~ this]
```

src/urbit/app/chess.hoon
```diff
@@ -676,51 +764,98 @@
...
+  ^-  [new=(unit active-game-state) cards=(list card)]
+  ?.  ?=(~ result.game.game)
     [~ ~]
   =/  new-position
-    (~(apply-move with-position position) move)
+    (~(apply-move with-position position.game) move)
   ?~  new-position
     [~ ~]
-  =/  updated-game  `chess-game`game
+  =/  updated-game  `chess-game`game.game
```
bonbud-macryg: Would it be reasonable to change a face somewhere to avoid instances of `game.game`?

src/urbit/app/chess.hoon
```diff
@@ -676,51 +764,98 @@
...
+              !>([%result game-id.game.game (need result.updated-game)])
           ==
-          :*  %give  %kick  :~  /game/(scot %da game-id.game)/updates
-                                /game/(scot %da game-id.game)/moves
+          :*  %give  %kick  :~  /game/(scot %da game-id.game.game)/updates
+                                /game/(scot %da game-id.game.game)/moves
                             ==
               ~
           ==
       ==
```
bonbud-macryg: Just indent this once, maybe also the branch below to keep this layout:
```
?.  <test>
  <if-false>
<if-true>
```

src/urbit/app/chess.hoon
```diff
@@ -676,51 +764,98 @@
...
+        (~(put by fen-repetition) fen 1)
+      (~(put by fen-repetition) fen +((need count)))
+  ++  check-threefold
+    |=  [fen-repetition=(map @t @ud) position=chess-position]
+      ^-  ?
+      =/  fen  ~(simplified position-to-fen position)
+      =*  count  (~(get by fen-repetition) fen)
+      ?~  count
+        |
+      (gth (need count) 2)
```
bonbud-macryg: Should outdent everything below the `|=` once in both arms.

src/urbit/mar/chess/result.hoon
```diff
@@ -0,0 +1,19 @@
+/+  chess
+=,  format
```
bonbud-macryg: It's a non-issue, but feel free to add a `=,  chess` below this line if you like. You can't just do `=,  chess, format`, which is odd.

src/urbit/mar/chess/update.hoon
```diff
@@ -40,6 +40,11 @@
         :~  ['chessUpdate' [%s 'draw-declined']]
             ['gameID' [%s (scot %da game-id.upd)]]
         ==
+      %draw-claim-available
+        %-  pairs:enjs
+        :~  ['chessUpdate' [%s 'draw-claim-available']]
+            ['gameID' [%s (scot %da game-id.upd)]]
+        ==
```
bonbud-macryg: As with the `special-draw` poke in `/app`, I think this should be something like `threefold-draw-claim-available` if I'm not missing something.

src/urbit/sur/chess.hoon
```diff
@@ -153,6 +151,8 @@
       [%offer-draw game-id=@dau]
       [%accept-draw game-id=@dau]
       [%decline-draw game-id=@dau]
+      [%change-special-draw-preference game-id=@dau setting=?]
+      [%claim-special-draw game-id=@dau]
```
bonbud-macryg, line 238: As above, might want to change.

src/urbit/sur/chess.hoon
```diff
@@ -161,6 +161,7 @@
       [%result game-id=@dau result=chess-result]
       [%draw-offer game-id=@dau]
       [%draw-declined game-id=@dau]
+      [%draw-claim-available game-id=@dau]
```
bonbud-macryg: As above, might want to change.

src/urbit/app/chess.hoon
rovmug-ticfyn: It's the irregular form of [centis](https://developers.urbit.org/reference/hoon/rune/cen#-centis) to change just that one wing of `u.game`.

src/urbit/app/chess.hoon
rovmug-ticfyn: Per [this comment](https://github.com/ashelkovnykov/urbit-chess/pull/40#discussion_r980612639), the same pokes and settings will eventually be used for the 50 move rule as well.

src/urbit/app/chess.hoon
```diff
@@ -52,7 +55,8 @@
   ^-  (quip card _this)
   =/  old-state  !<(versioned-state old-state-vase)
   ?-  -.old-state
-    %0  [~ this(state old-state)]
+    %1  [~ this(state old-state)]
+    %0  [~ this(state *state-1)]
   ==
```
bonbud-macryg, line 71: Notes on `on-load`:
```
++  on-load
  ::  XX: state transition is causing need-have
  ::      error: %load-failed
  ::      have %0, need %1
  ::      is it struggling with the vase?
  ::
  ::  what on-load does:
  ::  unpack with zapgal.
  ::  test which state version the imported
  ::  data has with the wuthep rune.
  ::  convert from old version to new.
  ::  load into the state of the subject.
  ::
  |=  old-state-vase=vase
  ::  col 3 on line below errors
  ^-  (quip card _this)
  ::  col 3 and 18 below errors
  =/  old-state  !<(versioned-state old-state-vase)
  ?-  -.old-state
    %1  [~ this(state old-state)]
    %0  [~ this(state *state-1)]
  ==
```

src/urbit/app/chess.hoon
```diff
@@ -144,6 +148,17 @@
...
+          =/  game  (~(get by games) game-id.action)
+          ?~  game
+            :_  this
+            =/  err
+              "no active game with id {<game-id.action>}"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          %=  this
+            games  (~(put by games) game-id.action u.game(auto-claim-special-draws setting.action))
+          ==
```
bonbud-macryg: Amended `%change-special-draw-preference`:
```
%change-special-draw-preference
  =/  game  (~(get by games) game-id.action)
  ?~  game
    :_  this
    =/  err
      "no active game with id {<game-id.action>}"
    :~  [%give %poke-ack `~[leaf+err]]
    ==
  :-  ~
  %=  this
    games  (~(put by games) game-id.action u.game(auto-claim-special-draws setting.action))
  ==
```

src/urbit/app/chess.hoon
bonbud-macryg: Amended `%claim-special-draw`:
```
%claim-special-draw
  =/  game-state  (~(get by games) game-id.action)
  ?~  game-state
    :_  this
    =/  err
      "no active game with id {<game-id.action>}"
    :~  [%give %poke-ack `~[leaf+err]]
    ==
  ?.  (check-threefold fen-repetition.u.game-state position.u.game-state)
    :_  this
    =/  err
      "no special draw available for game {<game-id.action>}"
      :~  [%give %poke-ack `~[leaf+err]]
      ==
  :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
              %chess-result  !>([game-id.action %'½–½' ~])
          ==
          :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
              %chess-update
              !>([%result game-id.action %'½–½'])
          ==
          :*  %give  %kick  :~  /game/(scot %da game-id.action)/updates
                                /game/(scot %da game-id.action)/moves
                            ==
              ~
          ==
      ==
  =/  updated-game  game.u.game-state
  =.  result.updated-game  `(unit chess-result)``%'½–½'
  %=  this
    games    (~(del by games) game-id.action)
    archive  (~(put by archive) game-id.action updated-game)
  ==
```

bonbud-macryg (dismissed):
As discussed, my amendments here move this issue onto a need-have error in the `on-load` arm.

src/urbit/app/chess.hoon
```diff
@@ -144,6 +148,17 @@
...
           ==
+        %change-special-draw-preference
+          =/  game  (~(get by games) game-id.action)
+          ?~  game
+            :_  this
+            =/  err
+              "no active game with id {<game-id.action>}"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          %=  this
```
ashelkovnykov, line 198: Missing cards list above this line (don't forget that arms must always return `(quip card _this)`)

src/urbit/lib/chess.hoon
```diff
@@ -1,4 +1,4 @@
-/-  chess
+/-  chess, historical
```
ashelkovnykov: `sur/historical` shouldn't be imported here, it should be imported into `app/chess`

src/urbit/app/chess.hoon
```diff
@@ -213,6 +228,39 @@
...
           ==
+        %claim-special-draw
+          =/  game-state  (~(get by games) game-id.action)
+          ?~  game-state
+            :_  this
+            =/  err
+              "no active game with id {<game-id.action>}"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          ?.  (check-threefold fen-repetition.game-state position.game-state)
```
ashelkovnykov: Should be `fen-repetition.u.game-state` and `position.u.game-state`

ashelkovnykov (changes requested):
The build is failing with a "have-need" error because the types of `chess-move` and `chess-update` were also changed, but the originals were not moved to `sur/historical` and so the "old" state refers to the new types.

The way that I found this was by comparing the "have" output and the "need" output side by side, after getting the lines and indentation to align. If you do this, you will notice that the "have" type has a `%end` value possible in moves, whereas the "need" type does not.

rovmug-ticfyn:
`%chess-result` is currently not working on the opponent's ship when a player claims a threefold draw

src/urbit/app/chess.hoon
```diff
@@ -7,17 +7,20 @@
 ::  define state structures
 |%
 +$  versioned-state
-  $%  state-0
+  $%  state-0.historical
```
ashelkovnykov: Let's avoid `.historical` by adding a namespace import `=,  historical` as we do with the `chess` types on line 5.

src/urbit/app/chess.hoon
ashelkovnykov: This was resolved by @rovmug-ticfyn after my comments on Nov 2

src/urbit/app/chess.hoon
```diff
@@ -778,56 +886,96 @@
...
+      (~(put by fen-repetition) fen 1)
+    (~(put by fen-repetition) fen +((need count)))
+  ++  check-threefold
+    |=  [fen-repetition=(map @t @ud) position=chess-position]
+    ^-  ?
+    =/  fen  ~(simplified position-to-fen position)
+    =*  count  (~(get by fen-repetition) fen)
+    ?~  count
+      |
+    (gth (need count) 2)
```
ashelkovnykov: Spacing is incorrect here - all of this should be indented by 2 less

src/urbit/sur/chess.hoon
```diff
@@ -247,6 +248,8 @@
...
+      [%special-draw-declined game-id=@dau]
```
ashelkovnykov: I'm really sorry to do this to you: now that I'm seeing how this part of the design interacts with the rest of the code, I think your initial idea was better: there should be a field in `active-game-state` called `special-draw-available` (or something like that) and we should set it instead of having these two updates.

On the frontend, I'm picturing a button in the left-hand game panel that's either enabled or disabled based on whether this value is true/false in the game state. If the user clicks it, it'll send the `%claim-special-draw` poke. Declining the special draws will be an implicit action: if the user makes a regular move, then the `special-draw-available` field will be recalculated based on whether a position is being repeated for a third time (again) or whether we've reached the 50-move rule (again).

Sorry!

src/urbit/sur/chess.hoon
```diff
@@ -236,6 +234,9 @@
       [%offer-draw game-id=@dau]
       [%accept-draw game-id=@dau]
       [%decline-draw game-id=@dau]
+      [%change-special-draw-preference game-id=@dau setting=?]
+      [%claim-special-draw game-id=@dau]
+      [%decline-special-draw game-id=@dau]
```
ashelkovnykov: See above comment: we can get rid of this poke by making it implicit

src/urbit/app/chess.hoon
```diff
@@ -257,6 +273,53 @@
...
+            =/  err
+              "no active game with id {<game-id.action>}"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
+                      %chess-update
+                      !>([%special-draw-declined game-id.action])
+                  ==
+              ==
+          this
```
ashelkovnykov: See above comment: we can remove this poke by making it implicit

src/urbit/app/chess.hoon
```diff
@@ -257,6 +273,53 @@
...
           ==
+        %claim-special-draw
+          =/  game-state  (~(get by games) game-id.action)
+          ?~  game-state
+            :_  this
+            =/  err
+              "no active game with id {<game-id.action>}"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          ?.  (check-threefold fen-repetition.u.game-state position.u.game-state)
```
ashelkovnykov: Need another check below this one for the 50-move rule

src/urbit/sur/chess.hoon
```diff
@@ -265,4 +268,9 @@
       her-hash=(unit @uvH)
       revealed=_|
   ==
++$  chess-result-action
```
ashelkovnykov: I think `chess-game-result` would be a much better name

src/urbit/mar/chess/result.hoon
```diff
@@ -0,0 +1,12 @@
...
+++  grab
+  |%
+  ++  noun  chess-result-action:chess
+  --
+++  grow
+  |%
+  ++  noun  act
+  --
+++  grad  %noun
+--
```
ashelkovnykov: As is, I believe this file's name needs to be "result-action.hoon" to get picked up correctly. Taking the above comment into account, it would need to become "game-result.hoon".

src/urbit/mar/chess/update.hoon
```diff
@@ -40,6 +40,16 @@
...
+      %special-draw-declined
+        %-  pairs:enjs
+        :~  ['chessUpdate' [%s 'special-draw-declined']]
+            ['gameID' [%s (scot %da game-id.upd)]]
+        ==
```
ashelkovnykov: See above comments: let's get rid of these updates and represent the availability of a special draw as part of the `active-game-state`

src/urbit/mar/chess/action.hoon
```diff
@@ -48,6 +48,25 @@
...
+        %.  jon
+        %-  ot:dejs
+        :~  [%game-id (se:dejs %da)]
+        ==
+      %decline-special-draw
+        :-  %decline-special-draw
+        %.  jon
+        %-  ot:dejs
+        :~  [%game-id (se:dejs %da)]
+        ==
```
ashelkovnykov, line 69: See above comments: we can get rid of the decline poke by making a decline action implicit

src/urbit/app/chess.hoon
```diff
@@ -257,6 +273,53 @@
...
+            :_  this
+            =/  err
+              "no special draw available for game {<game-id.action>}"
+              :~  [%give %poke-ack `~[leaf+err]]
+              ==
+          :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
+                      %chess-result  !>([game-id.action %'½–½' ~])
```
ashelkovnykov: This needs to be `%chess-result-action` (or `%chess-game-result`, pending changes from above comments), not `%chess-result`

src/urbit/app/chess.hoon
```diff
@@ -301,14 +364,7 @@
...
           =,  u.new.move-result
-          :-  :_  cards.move-result
-              ::  send move to opponent
-              :*  %give
-                  %fact
-                  ~[/game/(scot %da game-id.action)/moves]
-                  %chess-move
-                  !>(move.action)
-              ==
+          :-  cards.move-result
```
ashelkovnykov: Your rebase undid my bug-fix here, need to add it back in

src/urbit/app/chess.hoon
```diff
@@ -778,56 +886,96 @@
...
+    (~(put by fen-repetition) fen +((need count)))
+  ++  check-threefold
+    |=  [fen-repetition=(map @t @ud) position=chess-position]
+    ^-  ?
+    =/  fen  ~(simplified position-to-fen position)
+    =*  count  (~(get by fen-repetition) fen)
+    ?~  count
+      |
+    (gth (need count) 2) 
 --
```
ashelkovnykov, line 973: Need to add another helper: `check-50-move-rule`

src/urbit/app/chess.hoon
```diff
@@ -750,12 +806,19 @@
...
               =,  u.new.move-result
-              :-  cards.move-result
+              :-  ?.  (check-threefold fen-repetition position)
+                    cards.move-result
+                  %+  weld  cards.move-result
+                  ^-  (list card)
+                  :~  :*  %give  %fact  ~[/game/(scot %da u.game-id)/updates]
+                          %chess-update  !>([%draw-claim-available u.game-id])
+                      ==
+                  ==
```
ashelkovnykov: This part is written correctly, but quite ugly (out of necessity) - more evidence for why I think we should just move special draws being available into `active-game-state`

src/urbit/app/chess.hoon
```diff
@@ -765,7 +828,52 @@
               ::  add new games to our list
               %=  this
                 games  %+  ~(put by games)  u.game-id
-                       [game position |2.u.game-state]
+                       u.new.move-result
+              ==
+            %chess-result
+              =/  result
+                !<([game-id=@dau result=chess-result move=(unit chess-move)] q.cage.sign)
```
ashelkovnykov: You shouldn't need to spell out the type manually, you should be able to use the existing type from `sur/chess.hoon`:
```
=/  result  !<(chess-game-result q.cage.sign)
```

src/urbit/app/chess.hoon
```diff
@@ -765,7 +828,52 @@
               ::  add new games to our list
               %=  this
                 games  %+  ~(put by games)  u.game-id
-                       [game position |2.u.game-state]
+                       u.new.move-result
+              ==
+            %chess-result
```
ashelkovnykov: See above comments: this should be `%chess-result-action`, which should be renamed to `%chess-game-result`

src/urbit/app/chess.hoon
```diff
@@ -765,7 +828,52 @@
...
+              =/  result
+                !<([game-id=@dau result=chess-result move=(unit chess-move)] q.cage.sign)
+              ?~  move.result
+                ?.  (check-threefold fen-repetition.u.game-state position.u.game-state)
+                  [~ this]  ::  nice try, cheater
```
ashelkovnykov: Need another check for 50-move rule after this one

src/urbit/app/chess.hoon
```diff
@@ -765,7 +828,52 @@
...
+                  archive  (~(put by archive) u.game-id updated-game)
+                ==
+              =/  move-result
+                (try-move u.game-state (need move.result))
+              ?~  new.move-result
+                [cards.move-result this]  ::  nice try, cheater    
+              =,  u.new.move-result
+              ?.  (check-threefold fen-repetition position)
+                [~ this]  ::  nice try, cheater
+              :-  %+  weld  cards.move-result
```
ashelkovnykov: We don't actually need the cards from `try-move`: the ones produced here are enough. When we archive the game, we just need to use the version of the game state returned to `new.move-result`, and update the `result` field of that `chess-game` object to the result sent to our agent in the update.

src/urbit/app/chess.hoon
```diff
@@ -765,7 +828,52 @@
...
+                      ==
+                      :*  %give  %kick  :~  /game/(scot %da u.game-id)/updates
+                                            /game/(scot %da u.game-id)/moves
+                                        ==
+                          ~
+                      ==
+                  ==
+              %=  this
+                games    (~(del by games) u.game-id)
+                archive  (~(put by archive) u.game-id game)
```
ashelkovnykov: The indentation is a little weird here, and we can do this without pinning to the subject. We can also avoid having to duplicate the code for the cases where we do or do not have a move in the result update. I'd change this to:
```
              =/  result-game-state
                ?~  move.result
                  u.game-state
                =/  move-result  (try-move u.game-state (need move.result))
                ?~  new.move-result
                  u.game-state
                u.new.move-result
              =,  result-game-state
              ?.  (check-threefold fen-repetition position)
                [~ this]  ::  nice try, cheater
              :_  %=  this
                    games    (~(del by games) u.game-id)
                    archive  (~(put by archive) u.game-id game.u.game-state(result `result.result))
                  ==
              :~  :*  %give  %fact  ~[/game/(scot %da u.game-id)/updates]
                      %chess-update
                      !>([%result u.game-id result.result])
                  ==
                  :*  %give  %kick  :~  /game/(scot %da u.game-id)/updates
                                        /game/(scot %da u.game-id)/moves
                                    ==
                      ~
                  ==
              ==
```
Technically, this will allow a special draw claim with an invalid move, but only in situations where the opponent already has a draw claim available, so it doesn't compromise the security of the game (but should be documented!)

src/urbit/app/chess.hoon
```diff
@@ -778,56 +886,96 @@
...
             %white  `%'0-1'
             %black  `%'1-0'
           ==
         !!
       ::  give a card of the game result to opponent ship
-      :-  `[updated-game u.new-position]
+      :-  `[updated-game u.new-position new-fen-repetition |3.game]
+      ?.  threefold-draw
+        :~  position-update-card
+            move-card
```
ashelkovnykov: I think there's a bug here: I think we're emitting `move-card` regardless of whether we're verifying our own move or the opponents

src/urbit/app/chess.hoon
```diff
@@ -778,56 +886,96 @@
...
+                                  /game/(scot %da game-id.game.game)/moves
+                              ==
+                ~
+            ==
+        ==
+  ::  inform opponent of new position
       :~  position-update-card
-          :*  %give  %fact  ~[/game/(scot %da game-id.game)/updates]
+          :*  %give  %fact  ~[/game/(scot %da game-id.game.game)/moves]
+              %chess-result
```
ashelkovnykov: This should be `%chess-result-action` (or `%chess-game-result`, pending above comments)

src/urbit/app/chess.hoon
ashelkovnykov: Agreed, `game` in the input to `try-move` should be `game-state`

ashelkovnykov (changes requested):
I left some comments that have to do with #5. You can ignore those for now, and add them in as a separate commit on this PR after the initial commit for the threefold draw is done. We can also add another commit to this PR for #3.

src/urbit/app/chess.hoon
ashelkovnykov: Yes, I asked for this change so that we could handle both threefold repetition and the 50-move rule with the same code

src/urbit/mar/chess/update.hoon
ashelkovnykov: See above - this was my call for merging threefold repetition and the 50-move rule into one logic flow

src/urbit/sur/chess.hoon
ashelkovnykov: See above - this was my call for merging threefold repetition and the 50-move rule into one logic flow

src/urbit/sur/chess.hoon
ashelkovnykov: See above - this was my call for merging threefold repetition and the 50-move rule into one logic flow

src/urbit/app/chess.hoon
rovmug-ticfyn: I moved the move card to `try-move` in order to avoid sending an update on the `/moves` wire when we're sending a `%chess-game-result`, which includes the move if needed to verify that a special draw is claimable. But I think you're right that my version is going to send the move whether we're making a move or verifying an opponent's move. Is it a problem if we end up sending two updates on the `/moves` wire—the move as well as a `%chess-game-result`? In [this comment](https://github.com/thecommons-urbit/chess/pull/40#pullrequestreview-1121057259) you mentioned that would technically break the logic of the draw claim being sent with a move. I guess the alternative is to add a check in this arm whether we're auto-claiming a special draw so that we don't send the two separate cards.

ashelkovnykov: Ah, yeah, I see - you're right. I guess we add another `ship-to-move` copy into `try-move` and use it to determine which cards to send? This'll be ugly but it's not so bad, since #44 also adds `ship-to-move` into `try-move` if I remember correctly.

src/urbit/lib/chess.hoon
rovmug-ticfyn: It seems like `sur` files have to be imported into `lib`. When I add the `/-` in `app/chess` it doesn't compile.

src/urbit/app/chess.hoon
rovmug-ticfyn: I believe there's a bug in the version you've written here that will archive a game without the last move when that move is sent with the `%chess-game-result`. The archive line should save the `game` from `result-game-state` rather than `u.game-state`.
```archive  (~(put by archive) u.game-id game(result `result.result))```

File: src/urbit/lib/chess.hoon (hunk 34 above)
Comment (ashelkovnykov) [reply src/urbit/lib/chess.hoon]: It absolutely compiles, I just tested it. "It doesn't compile" is unclear and vague: what exactly is the error?

If it's a syntax error about the `/-` rune, then make sure that the import is before the `/+` rune: the `fas` runes have a strict ordering (the order in which they appear in the [documentation](https://developers.urbit.org/reference/hoon/rune/fas)).

If it's a `-find` error, then make sure you're either:
1. Referencing the correct namespace for the type (e.g. `state-0.historical`)
2. Implicitly referencing the correct namespace by stripping it during import (e.g. `/-  *historical`)
3. Implicitly referencing the correct namespace by explicitly stripping it (e.g. `=,  historical`)

File: src/urbit/app/chess.hoon (hunk 54 above)
Comment (ashelkovnykov) [reply src/urbit/app/chess.hoon]: Yes, you're correct - good catch!

File: src/urbit/lib/chess.hoon (hunk 34 above)
Comment (rovmug-ticfyn) [reply src/urbit/lib/chess.hoon]: Thanks, I didn't know about the ordering for `fas` runes so I was getting a syntax error from the `/-`when I added it after the `/+`.

For the sake of my understanding, importing `sur/historical` into `lib` seemed to work and make those types accessible in `app`, but I assume it's better for style and clarity to import the file directly where it is being referenced?

Comment (ashelkovnykov) [reply src/urbit/lib/chess.hoon]: Yeah, I think that would be best - the only thing that complicates it is that I'm not sure if the compiler has any sort of import deduplication. For example, if we imported `historical` into both `lib/chess` and `app/chess`, would we have two copies of everything in our subject? This is important, because in another language, we would import `sur/chess` into both `lib/chess` and `app/chess`, trusting that duplicates would be filtered. Instead, we have to have the more complicated import rule of "import files wherever they're directly used, unless another file we're importing has already imported them."

File: src/urbit/sur/chess.hoon (hunk 38 above)
Comment (rovmug-ticfyn) [reply src/urbit/sur/chess.hoon]: Want to check what I'm thinking before I start writing it:

1. I assume that the true/false value of `special-draw-available` should be sent to the frontend with each position update. 
2. To claim the draw while making a move, a player still needs to have their preference set to auto-claim. To claim it when receiving a threefold repetition (or >50th move) a player could auto-claim or click that button and send the `%claim-special-draw` poke. 
3. That means that `%claim-special-draw` needs to check whose turn it is because it can only be claimed on a player's own turn. A player should not be able to make a repeated move without auto-claiming the draw, then subsequently claim the draw before their opponent makes a move.

**Comment by rovmug-ticfyn** [comment]
Backend should be ready for threefold draw at this point. Need to add separate commits for 50 move rule and resigning, then connect it all to the new frontend.

File: src/urbit/sur/chess.hoon (hunk 38 above)
Comment (ashelkovnykov) [reply src/urbit/sur/chess.hoon]: Yes, that's all correct

File: src/urbit/app/chess.hoon [hunk 57]
```diff
@@ -574,7 +643,7 @@
         %=  this
           challenges-sent  (~(del by challenges-sent) src.bowl)
           rng-state  (~(del by rng-state) src.bowl)
-          games  (~(put by games) u.game-id [new-game *chess-position & | |])
+          games  (~(put by games) u.game-id [new-game *chess-position *(map @t @ud) & | | | |])
```
ashelkovnykov: Bug: `special-draw-available` starts as true

src/urbit/app/chess.hoon
```diff
@@ -607,7 +676,7 @@
           ==
       %=  this
         challenges-sent  (~(del by challenges-sent) src.bowl)
-        games  (~(put by games) u.game-id [new-game *chess-position & | |])
+        games  (~(put by games) u.game-id [new-game *chess-position *(map @t @ud) & | | | |])
```
ashelkovnykov: Bug: `special-draw-available` starts as true

src/urbit/app/chess.hoon
```diff
@@ -765,7 +834,35 @@
...
+                       u.new.move-result
+              ==
+            %chess-game-result
+              =/  result  !<(chess-game-result q.cage.sign)
+              =/  result-game-state
+                ?~  move.result
+                  u.game-state
+                =/  move-result  (try-move u.game-state (need move.result))
+                ?~  new.move-result
+                  u.game-state
```
ashelkovnykov, line 855: Should add a comment here that technically this allows an opponent to claim a special draw with an invalid move, but only in situations where he can claim a special draw with no move, so it doesn't break the game's correctness

src/urbit/app/chess.hoon
```diff
@@ -301,14 +367,17 @@
...
+                    ::  don't send extra move card if auto-claiming special draw
+                    cards.move-result
+                  ::  send move to opponent
+                  :_  cards.move-result
+                  :*  %give
+                      %fact
+                      ~[/game/(scot %da game-id.action)/moves]
+                      %chess-move
+                      !>(move.action)
+                  ==
```
ashelkovnykov: Indentation is off here: these lines should be indented by 4 fewer spaces

src/urbit/app/chess.hoon
```diff
@@ -451,11 +520,11 @@
         ==
       =/  fen  (position-to-fen u.new-position)
       :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
-                  %chess-update  !>([%position game-id.action fen])
+                  %chess-update  !>([%position game-id.action fen |])
```
ashelkovnykov, line 529: Technically, the 50-move rule draw may be available, depending on the FEN

src/urbit/app/chess.hoon
```diff
@@ -451,11 +520,11 @@
...
               ==
           ==
       %=  this
-        games  (~(put by games) game-id.action [game.action u.new-position & | |])
+        games  (~(put by games) game-id.action [game.action u.new-position *(map @t @ud) & | | | |])
```
ashelkovnykov: `special-draw-available` may be true or false here, depending on the the injected FEN (as mentioned above). For now, it should be hard-coded to false, to match the above state.

src/urbit/sur/historic.hoon
```diff
@@ -0,0 +1,63 @@
...
+      [%move game-id=@dau move=chess-move-0]
+  ==
++$  chess-update-0
+  $%  [%challenge who=ship challenge=chess-challenge]
+      [%position game-id=@dau position=@t]
+      [%result game-id=@dau result=chess-result]
+      [%draw-offer game-id=@dau]
+      [%draw-declined game-id=@dau]
+  ==
+--
```
ashelkovnykov: Need newline at the end of the file

src/frontend/ts/types/urbitChess.ts
```diff
@@ -32,7 +32,9 @@ export enum Update {
   Position = 'position',
   Result = 'result',
   DrawOffer = 'draw-offer',
-  DrawDeclined = 'draw-declined'
+  DrawDeclined = 'draw-declined',
+  DrawClaimAvailable = 'draw-claim-available',
+  SpecialDrawDeclined = 'special-draw-declined'
```
ashelkovnykov: Don't need `SpecialDrawDeclined` - making a move instead of taking the draw is an implicit decline

src/frontend/ts/types/urbitChess.ts
```diff
@@ -42,7 +44,10 @@ export enum Action {
   OfferDraw = 'offer-draw',
   AcceptDraw = 'accept-draw',
   DeclineDraw = 'decline-draw',
-  Move = 'move'
+  Move = 'move',
+  ChangeSpecialDrawPreference = 'change-special-draw-preference',
+  ClaimSpecialDraw = 'claim-special-draw',
+  DeclineSpecialDraw = 'decline-special-draw'
```
ashelkovnykov: Don't need `DeclineSpecialDraw` - making a move instead of taking the draw is an implicit decline

src/frontend/ts/types/urbitChess.ts
```diff
@@ -32,7 +32,9 @@ export enum Update {
   Position = 'position',
   Result = 'result',
   DrawOffer = 'draw-offer',
-  DrawDeclined = 'draw-declined'
+  DrawDeclined = 'draw-declined',
+  DrawClaimAvailable = 'draw-claim-available',
```
ashelkovnykov: `DrawClaimAvailable` isn't used, since the special draw status is sent as a part of the position update

src/frontend/ts/types/urbitChess.ts
```diff
@@ -128,6 +134,16 @@ export interface DrawDeclinedUpdate extends ChessUpdate {
...
+export interface DrawClaimAvailableUpdate extends ChessUpdate {
+  chessUpdate: Update.DrawClaimAvailable
+  gameID: GameID
+}
+
+export interface SpecialDrawDeclinedUpdate extends ChessUpdate {
+  chessUpdate: Update.SpecialDrawDeclined
+  gameID: GameID
+}
+
```
ashelkovnykov: Both of these unused (see above)

src/frontend/ts/types/urbitChess.ts
```diff
@@ -167,6 +183,19 @@ export interface DeclineDrawAction extends ChessGameAction {
...
+export interface ClaimSpecialDrawAction extends ChessGameAction {
+  'chess-action': Action.ClaimSpecialDraw
+}
+
+export interface DeclineSpecialDrawAction extends ChessGameAction {
+  'chess-action': Action.DeclineSpecialDraw
+}
+
```
ashelkovnykov: This action unused (see above)

src/frontend/ts/helpers/urbitChess.ts
```diff
@@ -114,3 +124,21 @@ export function declineDraw (gameId: GameID): DeclineDrawAction {
...
+}
+
+export function declineSpecialDraw (gameId: GameID): DeclineSpecialDrawAction {
+  const action: DeclineSpecialDrawAction = {
+    'chess-action': Action.DeclineSpecialDraw,
+    'game-id': gameId
+  }
+
+  return action
+}
```
ashelkovnykov: Can be removed

src/frontend/ts/state/chessStore.ts
```diff
@@ -114,6 +118,44 @@ const useChessStore = create<ChessState>((set, get) => ({
...
+          drawClaimAvailable: false,
+          info: currentGame.info
+        }
+
+        set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
+        updateDisplayGame(updatedGame)
+
+        console.log('RECEIVED SPECIAL DRAW DECLINED UPDATE')
+        break
+      }
```
ashelkovnykov: Can be removed

src/frontend/tsx/Chessboard.tsx
```diff
@@ -275,6 +275,16 @@ export function Chessboard () {
...
+  const acceptSpecialDrawOnClick = async () => {
+    const gameID = displayGame.info.gameID
+    await pokeAction(urbit, claimSpecialDraw(gameID))
+  }
+
+  const declineSpecialDrawOnClick = async () => {
+    const gameID = displayGame.info.gameID
+    await pokeAction(urbit, declineSpecialDraw(gameID))
+  }
+
```
ashelkovnykov: Can be removed

src/frontend/tsx/Chessboard.tsx
```diff
@@ -374,6 +384,21 @@ export function Chessboard () {
...
+          <br/>
+          <div className='draw-resolution row'>
+            <button className="accept" role="button" onClick={acceptSpecialDrawOnClick}>accept</button>
+            <button className="reject" role="button" onClick={declineSpecialDrawOnClick}>decline</button>
+          </div>
+        </div>
+      </Popup>
+    )
+  }
+
```
ashelkovnykov: I don't think this should be a pop-up. I think it should be a button in the game panel that's deactivated unless a special draw is available: if I remember correctly, that's how lichess and chess.com handle it when auto-accept-draw is disabled.

src/frontend/tsx/Games.tsx
```diff
@@ -22,6 +22,15 @@ export function Games () {
     await pokeAction(urbit, offerDraw(gameID), null, () => { offeredDraw(gameID) })
   }
 
+  const [checked, setChecked] = useState(false)
```
ashelkovnykov: 1. Need a better name than just "checked". What this value really represents is the status of the user's preference for auto-claiming special draws, so it should reflect that.
2. There's a bug here right now, since this state is enabled/disabled per-game, but the state is represented by a single boolean. What we would need is a `set` in the frontend called something like "auto-claim-games", and we would add/remove games to/from it when the user preference changes. When rendering the checkbox, we would check for the current game ID in this set to know whether or not it should be rendered as checked.
3. It's important to note that for ultimate correctness, the frontend should add/remove the game from the set of games with auto-claim enabled *based on an update from the backend, not on click in the checkbox*. This is because the backend state in Urbit is the ultimate source of truth, and we don't want the frontend to get out of sync.

The above means a flow that looks something like this:
1. New game starts with default preference for auto-claiming special draws
2. Game info sent to frontend; add game ID to set of games with auto-claim enabled if necessary
3. When rendering game, compute checkbox state by checking if game ID in set
4. Checkbox function = send auto-claim preference poke with opposite of current setting
5. Backend receives this poke, updates state, and sends update to frontend with new preference
6. Frontend receives update, adds/removes game ID to/from set of set of games with auto-claim enabled

This means adding another member & function to `chessState`/`chessStore`, and it'll also mean adding another update type to the backend.

ashelkovnykov (changes requested):
Backend basically complete, though will require minor additions to match some feedback I've added for the frontend

src/frontend/tsx/Games.tsx
rovmug-ticfyn: Rather than the `set` of auto-claim-games, wouldn't it be more in line with how we currently store this kind of data to add `autoClaimSpecialDraws` to `ActiveGameInfo`?

ashelkovnykov: You're right, that would be better. In that case, the the frontend will need to update the state of automatic claims for the game independently. This isn't ideal, but we're already doing it with `receiveUpdate`, so we may as well be consistent.

src/frontend/tsx/Settings.tsx
```diff
@@ -1,13 +1,35 @@
...
         <h4 className="control-panel-header">Gameplay</h4>
+        {hasGame ? (
+           <label>
+           <input
+             type="checkbox"
+             checked={displayGame.autoClaimSpecialDraws}
+             onChange={handleCheckboxChange}
+           />
+           Auto-Claim Special Draws
+         </label>
```
ashelkovnykov: Indentation issues on these lines, the ts-linter complains about them

ashelkovnykov (approved):
One last error fix, and we're done. Tested locally - all threefold repetition cases handled correctly. Some GUI issues on the frontend, but nothing that warrants blocking merge. LGTM - well done!

PR #31, opened 2022-09-02
Add comments to app, lib, and sur files

bonbud-macryg:
Closes #28 apart from a couple of `TODO:`s, partially addresses #29 and #30.

PR #41, opened 2022-09-14
Implement backend to partially address #3

### LINKED ISSUE #3: Cannot resign if it's not your turn

bonbud-macryg:
See description at the [this commit](https://github.com/ashelkovnykov/urbit-chess/commit/2454822ad5622e51045f9917ad078050fe422da8) for what I'm sure/not sure about in the overall implementation, and desc. on the [`%end` move commit](https://github.com/ashelkovnykov/urbit-chess/commit/d6a4388bbfa02fa49f39799e065b264b6092a558) for a quick note on that.

### DIFF
--- a/src/urbit/app/chess.hoon
+++ b/src/urbit/app/chess.hoon
@@ -257,6 +257,43 @@
             ::  record that draw offer is gone
             games  (~(put by games) game-id.action u.game(got-draw-offer |))
           ==
+        %resign
+          =/  game-state  (~(get by games) game-id.action)
+          ::
+          ::  we received a resignation with a game id,
+          ::  do we have a game with that id?
+          ?~  game-state
+            ::  if not, error!
+            :_  this
+            =/  err
+              "no active game with id {<game-id.action>}"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          ::  else, is the given result a draw?
+          ?:  =(result.action %'½–½')
+            ::  if so, error!
+            :_  this
+            =/  err
+              "resignation doesn't end in a draw"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          ::  else, return (quip card _this)
+                  ::  tell our opponent we've resigned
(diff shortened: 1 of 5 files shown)


ashelkovnykov:
Will refrain from reviewing until after #31 is merged

src/urbit/app/chess.hoon
```diff
@@ -257,6 +257,43 @@
             ::  record that draw offer is gone
             games  (~(put by games) game-id.action u.game(got-draw-offer |))
           ==
+        %resign
+          =/  game-state  (~(get by games) game-id.action)
+          ::
+          ::  we received a resignation with a game id,
+          ::  do we have a game with that id?
+          ?~  game-state
+            ::  if not, error!
```
bonbud-macryg, line 266: As mentioned in the chat, I have a note to address the "old-style" comments in this PR in line with your previous reviews.

ashelkovnykov:
In a vacuum, this is a good backend commit for issue #3. However, the changes by @rovmug-ticfyn in #40 might also solve this issue in a way that would make the current design in this commit (`%resign`) obsolete. For now, let's leave this PR open and return to it once #40 is merged.

ashelkovnykov:
@bonbud-macryg The work in #40 is basically a superset of this PR, plus work for #4 and #5. Would you prefer:
- We close this PR and open a new one to implement #3 after #40 is done
- We keep this PR open and rebuild on top of #40
- We close this PR and @rovmug-ticfyn can add code for #3 into #40, you keep going with `%telos`

Your call, all are doable!

bonbud-macryg:
If @rovmug-ticfyn is happy to I'd say we close this PR and he can implement a version of this code in #40. When I was initially divvying up the issues I was partly basing it on which could be bundled together and felt like discrete projects, so by now he knows that part of the codebase very well.

rovmug-ticfyn:
Fine with me!

PR #43, opened 2022-09-25
Implement customizable visuals on the frontend

### LINKED ISSUE #9: Customizable board visuals
One of the reasons I chose Chessground for the board is that it has [many board visualization options](https://github.com/lichess-org/lila/tree/master/public/piece-css) which can be plugged in. Currently, only the default visuals are available to `%chess`. Allowing users to choose their own board visuals would be very cool.

Like #8 , we should store the user's preferences. I don't think the `%chess` app state is the appropriate place for this, as the backend should be agnostic to how people use it (e.g. both Landscape and EScape use `%graph-store`).

Frontend work only.

brbenji:
This partially resolves #9.

**Please look over:**
src/frontend/tsx/Chessboard.tsx
src/frontend/css/boards.css
src/frontend/css/pieces.css

Everything else is just assets for themes.

**A little background:**
CSS class names are how chessground swaps board and piece **themes**. For a cleaner import section of Chessground.tsx, I decided to put all the board themes in one CSS file and picked a handful of piece themes to go into another, _maybe there is a better way._ 

Right now, the only way to change themes is manually changing the boardTheme and pieceTheme variables in Chessboard.tsx.

Screen Shot 2022-09-25 at 9 10 16 AM


src/frontend/tsx/Chessboard.tsx
```diff
@@ -22,6 +22,17 @@ import { Side, CastleSide, PromotionRole, Rank, File, GameID, ActiveGameInfo } f
 import 'chessground/assets/chessground.base.css'
 import 'chessground/assets/chessground.brown.css'
 import 'chessground/assets/chessground.cburnett.css'
+import '../css/pieces.css'
+import '../css/boards.css'
```
ashelkovnykov: You shouldn't need to import local `.css` files into the `.tsx` files. You can just add them to the list of `.css` files to link in `index.html`.

I'm not sure if this is actually the "correct" way to do this, since I've never done frontend professionally, but it's the way the rest of the code is, so let's just be consistent for now. If it's an issue, it'll get resolved later when the frontend gets a professional makeover.

src/frontend/tsx/Chessboard.tsx
```diff
@@ -345,8 +356,10 @@ export function Chessboard () {
         </p>
       </div>
       <div className='board-container'>
-        <div ref={boardRef} className='chessboard cg-wrap' />
-        { (promotionMove !== null) ? renderPromotionInterface() : <div/> }
+        <div className={`${boardTheme} ${pieceTheme}`}>
```
ashelkovnykov: I think this is the correct way to go. The next step is to add an interface for the user to select which board and piece theme he wants and code to inject the appropriate class name into the `div` accordingly. If you want help with this, let me know. Otherwise, I give you free reign creatively, and we can discuss differences of opinion (if any exist) once the code is out.

src/frontend/css/boards.css
```diff
@@ -0,0 +1,46 @@
+/* svg */
+.blue cg-board {
+  background-image:url('data:image/svg+xml;base64,PD94bWwgdmVyc2lvbj0iMS4wIiBlbmNvZGluZz0iVVRGLTgiIHN0YW5kYWxvbmU9Im5vIj8+CjxzdmcgeG1sbnM9Imh0dHA6Ly93d3cudzMub3JnLzIwMDAvc3ZnIiB4bWxuczp4PSJodHRwOi8vd3d3LnczLm9yZy8xOTk5L3hsaW5rIgogICAgIHZpZXdCb3g9IjAgMCA4IDgiIHNoYXBlLXJlbmRlcmluZz0iY3Jpc3BFZGdlcyI+CjxnIGlkPSJhIj4KICA8ZyBpZD0iYiI+CiAgICA8ZyBpZD0iYyI+CiAgICAgIDxnIGlkPSJkIj4KICAgICAgICA8cmVjdCB3aWR0aD0iMSIgaGVpZ2h0PSIxIiBmaWxsPSIjZGVlM2U2IiBpZD0iZSIvPgogICAgICAgIDx1c2UgeD0iMSIgeT0iMSIgaHJlZj0iI2UiIHg6aHJlZj0iI2UiLz4KICAgICAgICA8cmVjdCB5PSIxIiB3aWR0aD0iMSIgaGVpZ2h0PSIxIiBmaWxsPSIjOGNhMmFkIiBpZD0iZiIvPgogICAgICAgIDx1c2UgeD0iMSIgeT0iLTEiIGhyZWY9IiNmIiB4OmhyZWY9IiNmIi8+CiAgICAgIDwvZz4KICAgICAgPHVzZSB4PSIyIiBocmVmPSIjZCIgeDpocmVmPSIjZCIvPgogICAgPC9nPgogICAgPHVzZSB4PSI0IiBocmVmPSIjYyIgeDpocmVmPSIjYyIvPgogIDwvZz4KICA8dXNlIHk9IjIiIGhyZWY9IiNiIiB4OmhyZWY9IiNiIi8+CjwvZz4KPHVzZSB5PSI0IiBocmVmPSIjYSIgeDpocmVmPSIjYSIvPgo8L3N2Zz4K')
+}
```
ashelkovnykov, line 7: Why do some of these use image files and others some weird SVG hash? Why can't they all just use image files?

src/frontend/css/pieces.css
```diff
@@ -0,0 +1,111 @@
...
...sLTgxLjU4NzI0IC03OS40MDE3NCwtMjg3Ljg1MDg3IDc5LjMzNjc1LC0xNDkuMTMyMjkgMzAyLjk5NTU1LC0xMTQuMzI2ODEgMzYxLjE4ODg1LDUwLjIwNDkxIDYzLjI1MDEsMTc4LjgyOTI2IC0xMTIuODE2NSwyNDAuMDU2NDUgLTExMi44MTY1LDI0MC4wNTY0NSBMIDEzMjQuOTQzLDk4Ni43ODk2MyBaIE0gNDk4LjY5MzE1LDE3NzkuODA1NCBjIDAsMCAtNy44NTI1MywtMTA2Ljg3MzMgNTAuODAxODUsLTE1NC4yNjA4IDE5NS40ODc1NiwtMTczLjAxNTggMzM3LjI4NjY1LC0zODUuMzcwNiAzNzAuODg3MjYsLTU5OC45MTY1IGwgMjA5LjIzNDE0LDEuOTkzNSBjIDMzLjM0ODQsMjA2LjU1MzcgMTgyLjUwMDYsNDQ3LjAwNDEgMzIxLjAzNzIsNTYxLjc0NzEgMTAxLjczNiw1OS4wNTI0IDk4Ljk4NDIsMTk4LjM5MzcgOTguOTg0MiwxOTguMzkzNyB6IiBmaWxsPSIjZjlmOWY5Ii8+PHBhdGggZD0iTTUyMCAxNzY5aDEwMDhxOC05Ny0xMzItMTgyLTEzMi0xMDEtMTk2LjUtMjM5LjVUMTEyMCAxMDM5SDkyOHEtMTUgMTcwLTc5LjUgMzA4LjVUNjUyIDE1ODdxLTE0MSA4NS0xMzIgMTgyem01MDQgNzRINDQ2di03NHEtNC04MCA0MS41LTEzN1Q2MTMgMTUyNHExMTctOTEgMTcxLjUtMjE3LjVUODYzIDEwMzlINTc2bDI4NC0yMzlxLTg2LTc0LTg2LTE4OCAwLTEwMyA3My0xNzd0MTc3LTc0cTEwMyAwIDE3Ni41IDc0dDczLjUgMTc3cTAgMTE0LTg2IDE4OGwyODQgMjM5aC0yODdxMjMgMTQxIDc4IDI2Ny41dDE3MiAyMTcuNXE3OSA1MSAxMjQuNSAxMDh0NDIuNSAxMzd2NzR6TTc1NiA5NzRoNTM2bC0yMjUtMTkxcTEzNC0zMSAxMzQtMTcxIDAtNzYtNTIuNS0xMjYuNVQxMDI0IDQzNXEtNzMgMC0xMjUgNTAuNVQ4NDcgNjEycTAgMTQwIDEzNCAxNzF6IiBmaWxsPSIjMTAxMDEwIi8+PC9zdmc+')}
```
ashelkovnykov: This whole file seems to be just redefining the CSS from the individual files in `../theme/piece`

ashelkovnykov (changes requested):
Did you take a look at the [Chessground examples repo](https://github.com/lichess-org/chessground-examples)? I was expecting something more like the `assets` dir from that repo: a bunch of board/piece SVGs organized into directories, and then one or two big CSS files to setup the themes:
```
/**
 * Board themes
 */

// Blue
.blue .cg-wrap {
  background-image: url('images/board/blue.svg');
}
...

/**
 * Piece themes
 */

// Merida
.merida .cg-wrap piece.pawn.white {
  background-image: url('images/pieces/merida/wP.svg');
}
.merida .cg-wrap piece.rook.white {
...
```

ashelkovnykov:
It's an open question about what the best way to save the user's preferences for board/piece state between sessions. On one hand, it would be nice to toss it into the `%chess` agent. On the other hand, linking the interface with the backend that closely is not desirable. Session state would mean that it's transient between browser reloads. Cookies means that it can be disabled. URL parameters is possible but seems kinda dumb.

@tomholford Any advice as a full-stack guy?

tomholford:
> It's an open question about what the best way to save the user's preferences for board/piece state between sessions. On one hand, it would be nice to toss it into the `%chess` agent. On the other hand, linking the interface with the backend that closely is not desirable. Session state would mean that it's transient between browser reloads. Cookies means that it can be disabled. URL parameters is possible but seems kinda dumb.
> 
> @tomholford Any advice as a full-stack guy?

Depends on the UX that y'all are going for. One simple option is [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) in the browser, which will persist across sessions and doesn't rely on URL params or disabled cookies. The only downside is that if a user accesses chess games through multiple devices, they will have to toggle the settings on each one.

Depending on how much needs to be set (is it just a `theme` value, or more involved?), extending the `%chess` agent state is probably the lowest lift. But agreed that it doesn't quite fit, and encourages further scope creep. As an alternative, could create another simple `%chess-settings` agent, or even use the built-in `%settings-store`.

ashelkovnykov:
> Depends on the UX that y'all are going for. One simple option is [localStorage](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage) in the browser, which will persist across sessions and doesn't rely on URL params or disabled cookies. The only downside is that if a user accesses chess games through multiple devices, they will have to toggle the settings on each one.
> 
> Depending on how much needs to be set (is it just a `theme` value, or more involved?), extending the `%chess` agent state is probably the lowest lift. But agreed that it doesn't quite fit, and encourages further scope creep. As an alternative, could create another simple `%chess-settings` agent, or even use the built-in `%settings-store`.

The number of settings is low, and most would remain defaults:
- board theme
- piece theme
- auto-claim draw preference
- ELO server
- Timing server

Getting `%settings-store` involved would be even worse than adding the state to `%chess` IMO, since now there'd be third-party agent involved. I think localStorage is the best course, thanks for the tip.

src/frontend/tsx/Chessboard.tsx
brbenji: To answer your question above: I did look over the chessground-examples repo. That helped me understand how to “select” themes with the CSS class names. But I was basing most of my decision on the defaults in [chessground](https://github.com/lichess-org/chessground) and how [lichess’ main repo](https://github.com/lichess-org/lila/tree/master/public) is setup.

You're probably right about the best way to import `.css` files. I’m not totally sure why the chessground defaults have to be imported into `.tsx`. I thought I read it somewhere, but it might have gotten that impression from this mysterious message in the [chessground-example/assets/chessground.css](https://github.com/lichess-org/chessground-examples/blob/master/assets/chessground.css)
```
/*
 * Chessground base css properties.
 *
 * You need to include the css files in themes folder in order to have the
 * board and pieces displayed!
 */

.cg-wrap {
  width: 320px;
  height: 320px;
  position: relative;
  display: block;
}
...
```
I don’t know what `themes folder` this is referencing.

src/frontend/css/pieces.css
brbenji: Yes. `../theme/piece` at this point is just holding all the possible pieces for convenience. This file is making this smaller collection of piece themes available to the chessground by appending a new class name.

src/frontend/css/boards.css
brbenji: [lichess pieces-css](https://github.com/lichess-org/lila/tree/master/public/piece-css) as well as chessground use this SVG as Base64 code. I assumed it was the slicker and potentially lighter-weight way of doing SVG. So I ran with it.

It all could be simple image files. When it comes to the boards with textures they have to be either PNG or JPG, so if we wanted to be totally consistent that would make sense.

I used this [SVG to Base64 converter](https://base64.guru/converter/encode/image/svg) to convert things back and forth.

src/frontend/tsx/Chessboard.tsx
brbenji: I expect to figure out how to get an interface for theme selection going. The code already has examples of using React for these things. There’s already a <Settings /> component, that I assume I’ll be putting this in. I’ll probably stick close to how lichess.org does it.

src/frontend/tsx/Chessboard.tsx
ashelkovnykov: I think in case of an organization/style conflict, we should go with the Chessground example repo - it seems much more tractable than the main lichess repo, which is a nightmare to navigate (except for professional frontend guys and people who actively work on it, I'm sure...).

Agreed re: "themes" folder - no idea what it's talking about, especially when the layout and code is so clear on its own...

src/frontend/css/boards.css
ashelkovnykov: Apparently, it's called [inlining images](https://www.argentinaindependent.com/2018/08/12/inline-images.html), and it's mostly a relic of the past.

TL;DR - It makes the images load with the page, all at once. Without doing this, it's possible for the background to load first, then the pieces. This makes the image a component of the webpage, so it only loads when everything is ready together.

ashelkovnykov: I can see that being useful for a chessboard - only displaying the board when everything is ready.

src/frontend/css/pieces.css
ashelkovnykov: Gotcha, understood

ashelkovnykov:
Just took another look. Thoughts below:
1. I noticed that we have two copies of all the CSS right now, since we're importing in both `index.html` and `index.tsx`. I just tested with CSS imports only in `index.tsx` and everything worked fine with the Docker build, but if I remember correctly, the local `npm` build didn't work with the CSS imports only in `index.html`. In that case, let's remove the CSS imports from `index.html` and keep them all in `index.tsx`. We may as well remove the default board/piece style from `Chessground.tsx` and move the base Chessground assets CSS import into `index.html` as well:
  ```
  ...
  import App from './App'
  
  import 'chessground/assets/chessground.base.css'
  
  import '../css/tabs.css'
  import '../css/text.css'
  import '../css/boards.css'
  import '../css/pieces.css'
  import '../css/popups.css'
  import '../css/buttons.css'
  import '../css/general.css'
  import '../css/graphics.css'
  import '../css/textboxes.css'
  import '../css/responsive.css'
  import '../css/chessground.css'
  
  declare global {
  ...
  ```
2. I'm willing to compromise on the board/piece visuals if you feel strongly about it. Narrow down the choices to let's say 9 of each (so that we have a nice 3x3 box for people to choose from in the settings): how does that sound?
3. That's fine if we can't load the SVGs as SVGs. I was thinking we would inline all of the board visuals anyway. I remember you found a SVG to Base64 inliner yourself. Here's one for [JPG to Base64 inline](https://codebeautify.org/jpg-to-base64-converter). I tested it out on a random `.jpg` board ("horsey") and it worked.
4. Though they're massive, I like the monolithic `boards.css` and `pieces.css`, so as soon as we narrow down the board and piece candidates and inline all of the appropriate images, let's get rid of `../src/themes/*`.
5. I noticed that selected tile, previous move, and checkmate visuals don't work with the custom boards.

bonbud-macryg:
1. I removed the `index.html` imports while taking a look at this PR locally. Can confirm just using `index.tsx` works with the webpack-native build.
2. On the number of options: I've written about the boards on urbit. I said "five" on the number of piece options, but I don't have strong opinions about that number.
3. I think that visual preferences should persist across devices. %chess isn't responsive to smaller screens, it would be a bit of a project to make it so, but when/if it becomes mobile-friendly I'd be surprised if visual preferences not "syncing" across devices wasn't seen as a bug.

brbenji:
@ashelkovnykov 
> 2. I'm willing to compromise on the board/piece visuals if you feel strongly about it. Narrow down the choices to let's say 9 of each (so that we have a nice 3x3 box for people to choose from in the settings): how does that sound?

Cool. I’ll make an image of several options and post it in the [design notes](web+urbitgraph://group/~walbex/chess-improvements-bounty-workspace/graph/~walbex/design-notes-1019/170141184505938119262929044249654591488) bonbud started. We can all agree on as many as possible, and hold a ballot vote if we need to make a decision on a theme to eliminate. 

> 3. That's fine if we can't load the SVGs as SVGs. I was thinking we would inline all of the board visuals anyway. I remember you found a SVG to Base64 inliner yourself. Here's one for [JPG to Base64 inline](https://codebeautify.org/jpg-to-base64-converter). I tested it out on a random `.jpg` board ("horsey") and it worked.
> 4. Though they're massive, I like the monolithic `boards.css` and `pieces.css`, so as soon as we narrow down the board and piece candidates and inline all of the appropriate images, let's get rid of `../src/themes/*`.

I’ll make a commit that homogenizes everything into Base64 in the two CSS files.

> 5. I noticed that selected tile, previous move, and checkmate visuals don't work with the custom boards.

This totally went under my radar. I’ll look into making the highlighting work for every board

brbenji:
@bonbud-macryg 
> 3. I think that visual preferences should persist across devices. %chess isn't responsive to smaller screens, it would be a bit of a project to make it so, but when/if it becomes mobile-friendly I'd be surprised if visual preferences not "syncing" across devices wasn't seen as a bug.

This seems totally right. Does that mean %chess needs to be aware of visual preferences?

brbenji:
## Sorry this is a bit messy
**Here are the commits that matter. They are pretty self explanatory.**
[Refine board and piece selection. Convert imgs to base64.](https://github.com/thecommons-urbit/chess/pull/43/commits/26aa42807cc51090edcdfd17ef6b7d72189f9b9a)
[Remove css imports from index.html](https://github.com/thecommons-urbit/chess/pull/43/commits/b15a1fce489e09f05d585e0a06f7db84622747f9)
[Remove themes folder](https://github.com/thecommons-urbit/chess/pull/43/commits/c6adee1d9cb1fa775a311ef49b7939fb3e1ab40f)
[Move chessground style import to index.tsx](https://github.com/thecommons-urbit/chess/pull/43/commits/a652fe5e774e6bca6cf9c99f6ba8352b5e0f236f)

Everything works in both a docker and no-docker build. 

### Next
We’re missing is a `red.svg` and our `ivory and warm black` board. 
We have extra themes that I can remove if we commit to a set. 
 We might need dedicated files for a thumbnail when picking a board, but I want to explore better ways of doing it than how lichess.org does it.

src/frontend/tsx/index.tsx
```diff
@@ -1,13 +1,17 @@
...
+import '../css/pieces.css'
 import '../css/popups.css'
 import '../css/buttons.css'
 import '../css/general.css'
 import '../css/graphics.css'
 import '../css/textboxes.css'
 import '../css/responsive.css'
 import '../css/chessground.css'
+import 'chessground/assets/chessground.base.css' // Import Chessground style sheets
+import 'chessground/assets/chessground.brown.css'
```
ashelkovnykov: We don't need this one anymore, right? Since we're going to use our own brown as the default.

src/frontend/tsx/Chessboard.tsx
```diff
@@ -16,12 +16,13 @@ import { PromotionMove } from '../ts/types/chessground'
 import { Side, CastleSide, PromotionRole, Rank, File, GameID, ActiveGameInfo } from '../ts/types/urbitChess'
 
 //
-// Import Chessground style sheets
+// Choose different themese here
+// pieces: alpha, california, cardinal, cburnett, tatiana, chess7, chessnut, governor, kosal, maestro
+// boards: brown, green, blue, purple, wood4, grey, marble, blue2, pink-pyramid
```
ashelkovnykov: This is out of date now, right?

ashelkovnykov (commented):
Holding back feedback pending update which includes code for changing the board/piece visuals in-game

bonbud-macryg:
On the topic of visuals persisting across devices: if we split the frontend and backend repo (as discussed in #84), the intention is that the backend desk could eventually be something like a shared standard which many frontends could build on top of.

With that in mind, we shouldn't store the visual settings in the `%chess` agent proper, and I agree we shouldn't rely on a third-party `%settings-store`. A middle road between those two options might be a somewhat self-contained `%chess-settings` agent which would, in the short term, contain the following:
* Board theme, which won't be shared across frontends (they might not use chessground, they might not use our custom boards)
* Piece theme, which won't be shared across frontends
* Elo server, which other apps might not want to use
* Timing server, which other apps might not want to use
* Auto-claim draw preference. Draw functionality will be common to all chess apps using the `%chess` backend, but that you'd want draws to be auto-claimed could be particular to `%chess`. A lot of our UX is very quick and responsive just because there's so little to it. (Getting a practice game set up is a zero-click process after you've opened the app. The UI is so small it all loads instantly. We've talked about one-click "quick challenges" for pals.) It makes sense that we'd want to save the player time wherever possible. Auto-claiming draws is an obvious fit for `%chess`, but it might not be right for someone else's app/frontend. It makes sense that this would be particular to `%chess` and our `%chess-settings`.

When a developer clones the `%chess` backend upon which they're building a frontend, they'll also be downloading the `%chess-settings` agent contained in it, but `%chess-settings` separates the settings particular to our app from the settings pretty much everyone would want in theirs. It would save developers having to comb through files to remove our settings, and give them a sensible place to store their own.

ashelkovnykov:
@bonbud-macryg I like the idea of a `%chess-store`. My first thought was that such an agent should be in the same repo as the frontend to which it links, since that's really whose settings it's storing. My next thought was that this would be annoying, because no one wants to have to install the frontend for their chess app separately.

However, my final thought (in this chain of thinking) was that:
1. If done as I described above, the `%chess` agent serves as a core and a protocol, and lives in its own repo. The frontend and its associated `%chess-store` are a visualization tool and live in a separate repo. Individual devs can combine these two repos into a chess app in a single repo using git submodules. Aside from the submodules, this third git repo would contain app assets (e.g. docket, tile image, etc.) and installation scripts/instructions for hosts. I think this is the simplest and most modular way for developers to use our work - especially if we break the existing `%chess` app into three repos and provide examples + an article explaining the flow in The Commons wiki.
2. The above flow neatly encapsulates the different responsibilities for devs, hosts, and users: devs write individual agents + frontends then combine them into apps; hosts follow the instructions to run apps on their ship; users install apps from a host in Landscape.

The only point on which I'm not certain I agree is about whether `auto-claim-special-draws` should be a `%chess`-level preference or not. I consider it core to a proper chess implementation/protocol, since it affects which legal actions a user is allowed to take. However, the way that lichess and chess.com handle it is closed to how you describe: a frontend detail.

src/frontend/ts/helpers/themes.ts
```diff
@@ -0,0 +1,18 @@
+//
+// Add different themes here
+// Here are a few options still in the code. Quickly see these by
+// changing the CSS class name in Dev mode in a browser.
+// pieces: alpha, california, cardinal, cburnett, tatiana, chess7, chessnut, governor, kosal, maestro
```
ashelkovnykov: Outdated comment for piece names

src/frontend/tsx/App.tsx
```diff
@@ -10,6 +10,14 @@ import { GamePanel } from './GamePanel'
 export function App () {
   const { urbit, setUrbit, receiveChallenge, receiveGame } = useChessStore()
 
+  //
+  // State
+  // XX: Make the state more efficient with one array (or something) for prefereneces.
+  // const [preferences, setPreferences] = useState([])
+  //
+  const [pieceTheme, setPieceTheme] = useState<string>('tatiana')
+  const [boardTheme, setBoardTheme] = useState<string>('blue')
```
ashelkovnykov: On its own, this state will reset on reload/refresh, meaning that the user needs to set his visual preferences every single time he opens the chess app. As discussed in the comments on this PR, you're going to want to combine the React state with [`localStorage`](https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage). [Here](https://www.freecodecamp.org/news/how-to-use-localstorage-with-react-hooks-to-set-and-get-items/) is an example guide, but you might want to find another one (there are several online).

src/frontend/tsx/App.tsx
```diff
@@ -55,8 +63,12 @@ export function App () {
...
       <div className='app-container'>
         <GamePanel />
-        <Chessboard />
-        <Menu />
+        <Chessboard pieceTheme={pieceTheme} boardTheme={boardTheme}/>
+        <Menu
+          pieceTheme={pieceTheme}
+          setPieceTheme={setPieceTheme}
+          setBoardTheme={setBoardTheme}
+        />
```
ashelkovnykov: I understand why you're doing this, since every single visual components needs one or both of the themes. However, I really dislike this method of passing info between the different visual components. Especially with `localStorage`, it shouldn't be necessary since we can directly load the themes in each component.

However, my dislike is not a sufficient reason to not use this tool. Furthermore, you're the one actively writing this code, so if you find that this is the easiest way to access/use the info, I'll let you make that executive decision.

src/frontend/tsx/Chessboard.tsx
```diff
@@ -42,7 +34,7 @@ declare global {
 // Main
 //
 
-export function Chessboard () {
+export function Chessboard ({pieceTheme, boardTheme}: any) {
```
ashelkovnykov: It's considered poor craftsmanship to use the 'any' type in this situation. If we're going to pass the theme info between components, the correct React way to do it would be like this:
```
type ChessboardProps = {
  pieceTheme: string,
  boardTheme: string
}

export function Chessboard ({pieceTheme, boardTheme}: ChessboardProps) {
  ...
```

src/frontend/tsx/Games.tsx
```diff
@@ -3,7 +3,7 @@ import { pokeAction, resign, offerDraw } from '../ts/helpers/urbitChess'
 import { Side, GameID, GameInfo, ActiveGameInfo } from '../ts/types/urbitChess'
 import useChessStore from '../ts/state/chessStore'
 
-export function Games () {
+export function Games ({pieceTheme, boardTheme}: any) {
```
ashelkovnykov: See above comment

src/frontend/tsx/Menu.tsx
```diff
@@ -4,7 +4,7 @@ import { Challenges } from './Challenges'
 import { Games } from './Games'
 import { Settings } from './Settings'
 
-export function Menu () {
+export function Menu ({pieceTheme, setPieceTheme, setBoardTheme}: any) {
```
ashelkovnykov: See above comment

src/frontend/tsx/Settings.tsx
```diff
@@ -1,10 +1,84 @@
...
+              }
+
+              return (
+                <li key={ind}
+                    id={theme}
+                    className={`theme ${theme}`}
+                    onClick={() => handleClick()}
+                >
+                  <piece className="theme-icon black knight"/>
+                </li>
```
ashelkovnykov: After combining `state` with `localStorage`, you shouldn't need to use `toggleSelect`. Just calling `setPieceTheme` should be enough, and the interface will rerender because the state of `localStorage` will change. Thus, you should be able to do something like this:
```
pieceThemes.map((theme, ind) => {
  
  status = (theme === localStorage.getItem('pieceTheme')) ? 'selected' : 'unselected'

  const handleClick = () => {
    setPieceTheme(`${theme}`)
  }

  return (
    <li key={ind}
        id={theme}
        className={`theme ${theme} ${status}`}
        onClick={() => handleClick()}
    >
      <piece className="theme-icon black knight"/>
    </li>)})
```

ashelkovnykov (changes requested):
Very close but needs a few changes:
- Need to rebase commits onto new master
- A lot of linter complaints
- Need to use `localStorage` (see comments)

src/frontend/ts/helpers/themes.ts
```diff
@@ -0,0 +1,10 @@
+export const pieceThemes:string[] = [
+  'tatiana', 'kosal', 'cardinal',
+  'maestro', 'alpha', 'california'
+]
+export const boardThemes:string[] = [
+  'blue', 'blue2', 'red',
+  'marble', 'default', 'purple',
+  'green', 'grey', 'brown',
+  'red2'
+]
```
ashelkovnykov: This file should go into `ts/constants` instead of `ts/helpers`

src/frontend/ts/state/preferenceState.ts
```diff
@@ -0,0 +1,8 @@
+interface PreferenceState {
+  pieceTheme: string | null;
+  boardTheme: string | null;
+  setPieceTheme: (theme: string) => void;
+  setBoardTheme: (theme: string) => void;
```
ashelkovnykov: Optional, but if we want to be safer, we could define the piece and board themes as enumerations (see `types/urbitChess`) and then require that the `PreferenceState` take a `BoardTheme`/`PieceTheme` enum. This would give us stricter typing, which is safer.

I think it's the right thing to do, but it doesn't have to go into this PR: we can just add it as an issue for now, and finish up this PR.

ashelkovnykov (commented):
Gorgeous. Great work with the visual design, but also with the `localStorage` - for what this PR does, the code is exceptionally clean and minimal.

My personal preference is `default` + `cardinal`

ashelkovnykov (changes requested):
LGTM. See above two comments and squash into a single commit (or however many you think make logical sense - the goal is to limit each commit to be a self-contained group of changes that build on top of one another), then we're good to merge.

If you want to squash to more than one commit, ["interactive rebase"](https://medium.com/swlh/tutorial-on-interactive-rebasing-e3f49505b2ce) is really helpful.

brbenji:
# Our new look!
Screen Shot 2022-12-12 at 5 16 43 PM

ashelkovnykov (approved):
LGTM

PR #45, opened 2022-09-25
Add scry path and wire for outgoing challenges

### LINKED ISSUE #7: Allow users to browse outgoing challenges
`%chess` stores outgoing challenges (obviously), but doesn't make them available in any way through Gall. As mentioned in #6 , we would need to think about what the best way to implement reading this data is, and then see if that's possible through the `http-api`. Will require both frontend and backend work.

bonbud-macryg:
Partially addresses #7.

As with the `%pals` integration, this code looks correct to me but I can't test that because I don't understand how the two marks in a scry interact.

Scry `.^(json %gx /=chess=/outgoing/json)` gets an error with `peek no tube from chess-outgoing to json`, if that's of use.

### DIFF
--- a/src/frontend/ts/state/chessState.ts
+++ b/src/frontend/ts/state/chessState.ts
@@ -2,20 +2,20 @@ import Urbit from '@urbit/http-api'
 import { Ship, GameID, GameInfo, ActiveGameInfo, Challenge, ChessUpdate, ChallengeUpdate } from '../types/urbitChess'
 
 interface ChessState {
-  urbit: Urbit | null;
-  displayGame: ActiveGameInfo | null;
-  practiceBoard: String | null;
-  activeGames: Map<GameID, ActiveGameInfo>;
-  incomingChallenges: Map<Ship, Challenge>;
-  setUrbit: (urbit: Urbit) => void;
-  setDisplayGame: (displayGame: ActiveGameInfo | null) => void;
-  setPracticeBoard: (practiceBoard: String | null) => void;
-  receiveChallenge: (data: ChallengeUpdate) => void;
-  receiveGame: (data: GameInfo) => void;
-  receiveUpdate: (data: ChessUpdate) => void;
-  removeChallenge: (who: Ship) => void;
-  declinedDraw: (gameID: GameID) => void;
-  offeredDraw: (gameID: GameID) => void;
+  urbit: Urbit | null
+  displayGame: ActiveGameInfo | null
+  practiceBoard: String | null
+  activeGames: Map<GameID, ActiveGameInfo>
+  incomingChallenges: Map<Ship, Challenge>
+  outgoingChallenges: Map<Ship, Challenge>
+  setUrbit: (urbit: Urbit) => void
+  setDisplayGame: (displayGame: ActiveGameInfo | null) => void
+  setPracticeBoard: (practiceBoard: String | null) => void
+  receiveChallengeUpdate: (data: ChallengeUpdate) => void
+  receiveGame: (data: GameInfo) => void
+  receiveUpdate: (data: ChessUpdate) => void
+  declinedDraw: (gameID: GameID) => void
+  offeredDraw: (gameID: GameID) => void
 }
 
 export default ChessState
--- a/src/frontend/ts/state/chessStore.ts
+++ b/src/frontend/ts/state/chessStore.ts
@@ -1,7 +1,7 @@
 import create from 'zustand'
 import Urbit from '@urbit/http-api'
 import { CHESS } from '../constants/chess'
-import { Update, Ship, GameID, GameInfo, ActiveGameInfo, Challenge, ChessUpdate, ChallengeUpdate, PositionUpdate, ResultUpdate, DrawOfferUpdate, DrawDeclinedUpdate, SpecialDrawPreferenceUpdate } from '../types/urbitChess'
+import { Update, Ship, GameID, GameInfo, ActiveGameInfo, Challenge, ChessUpdate, ChallengeUpdate, ChallengeSentUpdate, ChallengeReceivedUpdate, PositionUpdate, ResultUpdate, DrawOfferUpdate, DrawDeclinedUpdate, SpecialDrawPreferenceUpdate } from '../types/urbitChess'
 import ChessState from './chessState'
 
 const useChessStore = create<ChessState>((set, get) => ({
@@ -10,11 +10,42 @@ const useChessStore = create<ChessState>((set, get) => ({
   practiceBoard: '',
   activeGames: new Map(),
   incomingChallenges: new Map(),
+  outgoingChallenges: new Map(),
   setUrbit: (urbit: Urbit) => set({ urbit }),
   setDisplayGame: (displayGame: ActiveGameInfo | null) => set({ displayGame }),
   setPracticeBoard: (practiceBoard: String | null) => set({ practiceBoard }),
-  receiveChallenge: (data: ChallengeUpdate) =>
-    set(state => ({ incomingChallenges: state.incomingChallenges.set(data.who, data as Challenge) })),
+  receiveChallengeUpdate: (data: ChallengeUpdate) => {
+    switch (data.chessUpdate) {
+      case Update.ChallengeSent: {
+        set(state => ({ outgoingChallenges: state.outgoingChallenges.set(data.who, data as ChallengeSentUpdate) }))
+        break
+      }
+      case Update.ChallengeReceived: {
+        set(state => ({ incomingChallenges: state.incomingChallenges.set(data.who, data as ChallengeReceivedUpdate) }))
+        break
+      }
+      case Update.ChallengeResolved: {
+        let outgoingChallenges: Map<Ship, Challenge> = get().outgoingChallenges
+        outgoingChallenges.delete(data.who)
+
+        set({ outgoingChallenges })
+        break
+      }
+      case Update.ChallengeReplied: {
+        let incomingChallenges: Map<Ship, Challenge> = get().incomingChallenges
+        incomingChallenges.delete(data.who)
+
+        set({ incomingChallenges })
+        break
+      }
+      default: {
+        console.log('RECEIVED BAD UPDATE')
+        console.log(data.chessUpdate)
+        console.log((data as ChallengeUpdate).chessUpdate)
+        console.log((data as ChallengeUpdate).who)
+      }
+    }
+  },
   receiveGame: async (data: GameInfo) => {
     const activeGame: ActiveGameInfo = {
       position: CHESS.defaultFEN,
@@ -151,12 +182,6 @@ const useChessStore = create<ChessState>((set, get) => ({
         console.log((data as PositionUpdate).position)
       }
     }
-  },
-  removeChallenge: (who: Ship) => {
-    let incomingChallenges: Map<Ship, Challenge> = get().incomingChallenges
-    incomingChallenges.delete(who)
-
-    set({ incomingChallenges })
   },
   declinedDraw: (gameID: GameID) => {
     let updatedGame: ActiveGameInfo = get().activeGames.get(gameID)
--- a/src/frontend/ts/types/urbitChess.ts
+++ b/src/frontend/ts/types/urbitChess.ts
@@ -28,7 +28,10 @@ export enum Result {
 }
 
 export enum Update {
-  Challenge = 'challenge',
+  ChallengeSent = 'challenge-sent',
+  ChallengeReceived = 'challenge-received',
+  ChallengeResolved = 'challenge-resolved',
+  ChallengeReplied = 'challenge-replied',
   Position = 'position',
   Result = 'result',
   DrawOffer = 'draw-offer',
@@ -104,8 +107,22 @@ export interface ChessUpdate {
 }
 
 export interface ChallengeUpdate extends ChessUpdate {
-  chessUpdate: Update.Challenge
+  chessUpdate: Update.ChallengeSent |
+               Update.ChallengeReceived |
+               Update.ChallengeResolved |
+               Update.ChallengeReplied
   who: Ship
+}
+
+export interface ChallengeSentUpdate extends ChallengeUpdate {
+  chessUpdate: Update.ChallengeSent
+  challengerSide: Side
+  event: string
+  round: string
+}
+
+export interface ChallengeReceivedUpdate extends ChallengeUpdate {
+  chessUpdate: Update.ChallengeReceived
   challengerSide: Side
   event: string
   round: string
--- a/src/frontend/tsx/App.tsx
+++ b/src/frontend/tsx/App.tsx
@@ -8,7 +8,7 @@ import { Menu } from './Menu'
 import { GamePanel } from './GamePanel'
 
 export function App () {
-  const { urbit, setUrbit, receiveChallenge, receiveGame } = useChessStore()
+  const { urbit, setUrbit, receiveChallengeUpdate, receiveGame } = useChessStore()
 
   //
   // Helper functions
@@ -18,12 +18,11 @@ export function App () {
     const newUrbit = new Urbit('', '')
     newUrbit.ship = window.ship
     setUrbit(newUrbit)
-
     await newUrbit.subscribe({
       app: 'chess',
       path: '/challenges',
       err: () => {},
-      event: (data: ChallengeUpdate) => receiveChallenge(data),
+      event: (data: ChallengeUpdate) => receiveChallengeUpdate(data),
       quit: () => {}
     })
     await newUrbit.subscribe({
--- a/src/urbit/app/chess.hoon
+++ b/src/urbit/app/chess.hoon
@@ -90,8 +90,9 @@
               "already challenged {<who.action>}"
             :~  [%give %poke-ack `~[leaf+err]]
             ==
+          ::
           ::  send new challenge
-          :-  :~  :*  %pass  /poke/challenge  %agent  [who.action %chess]
+          :-  :~  :*  %pass  /poke/challenge/send  %agent  [who.action %chess]
                       %poke  %chess-challenge  !>(challenge.action)
                   ==
               ==
@@ -151,6 +152,10 @@
                   :*  %give  %fact  ~[/active-games]
                       %chess-game  !>(new-game)
                   ==
+                  ::  tell our frontend we accepted the challenge
+                  :*  %give  %fact   ~[/challenges]
+                      %chess-update  !>([%challenge-replied who.action])
+                  ==
               ==
           %=  this
             ::  remove our challenger from challenges-received
@@ -172,7 +177,7 @@
             :~  [%give %poke-ack `~[leaf+err]]
             ==
           ::  tell our challenger we decline
-          :-  :~  :*  %pass  /poke/challenge  %agent  [who.action %chess]
+          :-  :~  :*  %pass  /poke/challenge/reply  %agent  [who.action %chess]
                       %poke  %chess-decline-challenge  !>(~)
                   ==
               ==
@@ -402,7 +407,7 @@
     %chess-challenge
       =/  challenge  !<(chess-challenge vase)
       :-  :~  :*  %give  %fact  ~[/challenges]
-                  %chess-update  !>([%challenge src.bowl challenge])
+                  %chess-update  !>([%challenge-received src.bowl challenge])
               ==
           ==
       %=  this
@@ -412,7 +417,10 @@
     ::
     ::  handle declined challenges
     %chess-decline-challenge
-      :-  ~
+      :-  :~  :*  %give  %fact   ~[/challenges]
+                  %chess-update  !>([%challenge-resolved src.bowl])
+              ==
+          ==
       %=  this
         challenges-sent  (~(del by challenges-sent) src.bowl)
       ==
@@ -558,16 +566,27 @@
   ^-  (quip card _this)
   ?+  path  (on-watch:default path)
     ::
-    ::  convert incoming challenges to chess-update marks for subscribers
+    ::  get all challenge updates
     [%challenges ~]
       ?>  (team:title our.bowl src.bowl)
       :_  this
-      %+  turn  ~(tap by challenges-received)
-      |=  [who=ship challenge=chess-challenge]
-      ^-  card
-      :*  %give  %fact  ~
-          %chess-update  !>([%challenge who challenge])
-      ==
+      %-  zing
+      %-  limo
+      ::  challenges sent
+      :-  %+  turn  ~(tap by challenges-sent)
+          |=  [who=ship challenge=chess-challenge]
+          ^-  card
+          :*  %give  %fact   ~
+              %chess-update  !>([%challenge-sent who challenge])
+          ==
+      ::  challenges received
+      :-  %+  turn  ~(tap by challenges-received)
+          |=  [who=ship challenge=chess-challenge]
+          ^-  card
+          :*  %give  %fact   ~
+              %chess-update  !>([%challenge-received who challenge])
+          ==
+      ~
     ::
     ::  convert active games to chess-game marks for subscribers
     [%active-games ~]
@@ -645,6 +664,10 @@
                 :*  %give  %fact  ~[/active-games]
                     %chess-game  !>(new-game)
                 ==
+                ::  tell our frontend our challenge was accepted
+                :*  %give  %fact   ~[/challenges]
+                    %chess-update  !>([%challenge-resolved src.bowl])
+                ==
             ==
         %=  this
           challenges-sent  (~(del by challenges-sent) src.bowl)
@@ -679,11 +702,17 @@
               :*  %give  %fact  ~[/active-games]
                   %chess-game  !>(new-game)
               ==
+              ::  tell our frontend our challenge was accepted
+              :*  %give  %fact   ~[/challenges]
+                  %chess-update  !>([%challenge-resolved src.bowl])
+              ==
           ==
       %=  this
         challenges-sent  (~(del by challenges-sent) src.bowl)
         games  (~(put by games) u.game-id [new-game *chess-position *(map @t @ud) | & | | |])
       ==
+    ::
+    ::  XX: document this
     [%game @ta %updates ~]
       =/  game-id  `(unit @dau)`(slaw %da i.t.path)
       ?~  game-id
@@ -718,6 +747,7 @@
   ^-  (unit (unit cage))
   ?+  path  (on-peek:default path)
     ::
+    ::  .^(noun %gx /=chess=/game/~1996.2.16..10.00.00..0000/noun)
     ::  read game info
     ::  either active or archived
     [%x %game @ta ~]
@@ -730,15 +760,19 @@
         ``[%chess-game !>(u.archived-game)]
       ``[%chess-game !>(game.u.active-game)]
     ::
-    ::  .^(noun %gx /=chess=/friends/noun)
-    ::  .^(json %gx /=chess=/friends/json)
-    ::  read mutual friends
-    [%x %friends ~]
-      ``[%chess-pals !>((~(mutuals pals bowl) ~.))]
+    ::  .^(noun %gx /=chess=/challenges/outgoing/noun)
+    ::  list challenges sent
+    [%x %challenges %outgoing ~]
+      ``[%chess-challenges !>(~(tap by challenges-sent))]
     ::
-    ::  .^(arch %gy /=chess=/game)
+    ::  .^(noun %gx /=chess=/challenges/incoming/noun)
+    ::  list challenges received
+    [%x %challenges %incoming ~]
+      ``[%chess-challenges !>(~(tap by challenges-received))]
+    ::
+    ::  .^(arch %gy /=chess=/games)
     ::  collect all the game-id keys
-    [%y %game ~]
+    [%y %games ~]
       :-  ~  :-  ~
       :-  %arch
       !>  ^-  arch
@@ -749,6 +783,12 @@
       %+  turn  ids
       |=  a=@dau
       [(scot %da a) ~]
+    ::
+    ::  .^(noun %gx /=chess=/friends/noun)
+    ::  .^(json %gx /=chess=/friends/json)
+    ::  read mutual friends
+    [%x %friends ~]
+      ``[%chess-pals !>((~(mutuals pals bowl) ~.))]
   ==
 ++  on-agent
   |=  [=wire =sign:agent:gall]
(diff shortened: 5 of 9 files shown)


ashelkovnykov (changes requested):
When I wrote the issue for #7, I wrote down the first thing that came to mind for how to pull data out of the agent, which was a scry. Thinking further, that's a terrible idea and it should be an update on a subscriber wire like everything else. For now, that would mean adding a new field to update (two new fields?) for outgoing challenges to the `update` mark/wire. Eventually, the updates should be refactored into multiple different marks/wires so that you don't have incoming challenges, outgoing challenges, moves, and draws all in one place.

That said, I think having a scry path for incoming/outgoing challenges is a good idea. I think you should keep the existing scry, but rename it and split it into two: there should be `[%x %challenge %outgoing ~]` and `[%x %challenge %incoming ~]` (since both may be of interest). I'm thinking of refactoring the scries later to make all useful data fetchable out of `%chess` without wrapping it in `dbug`.

If you scroll up when you try your scry, you'll see the true error message. Hint: it's a problem with your mark. It's simple to correct, but with the above comments in mind, you'll need to make more changes to the mark. Luckily, you won't need to convert to JSON anymore.

Would you kindly please rebase your branch to the latest `master` to pick up the comments PRs, and then add example scries like `[%y %games ~]` currently has:
```
::  .^(chess-game %gx /=chess=/game/~2022.10.3..21.38.35..00d8.486c.24e9.9a71/noun)
::  .^((list challenge) %gx /=chess=/challenge/incoming/noun)
```

bonbud-macryg:
Current status: scry paths tested and work, haven't yet tested subscribing from the frontend.

src/urbit/app/chess.hoon
```diff
@@ -89,6 +89,10 @@
           :-  :~  :*  %pass  /poke/challenge  %agent  [who.action %chess]
                       %poke  %chess-challenge  !>(challenge.action)
                   ==
+                  ::  add new challenge to challenges-sent wire
+                  :*  %give  %fact   ~[/challenges-sent]
+                      %chess-update  !>([%chess-challenge who.action challenge.action])
```
ashelkovnykov: I believe this should be just `%challenge`, not `%chess-challenge`

src/urbit/app/chess.hoon
```diff
@@ -481,8 +485,11 @@
   |=  =path
   ^-  (quip card _this)
   ?+  path  (on-watch:default path)
-    ::
-    ::  convert incoming challenges to chess-update marks for subscribers
+  ::
+  ::  XX: should we rename this to
+  ::      %challenges-received for consistency's sake?
```
ashelkovnykov: I think that's a good idea - let's do that

src/urbit/app/chess.hoon
```diff
@@ -481,8 +485,11 @@
...
+  ::
+  ::  convert incoming challenges to chess-update marks for subscribers
```
ashelkovnykov: Old indentation was correct (same goes for all changes to comment indentation in this file)

src/urbit/app/chess.hoon
```diff
@@ -653,9 +674,22 @@
...
+    [%x %challenges %outgoing ~]
+      ``[%chess-outgoing !>(~(tap by challenges-sent))]
+  ::
+  ::  .^(noun %gx /=chess=/challenges/incoming/noun)
+  ::  list challenges received
+    [%x %challenges %incoming ~]
+      ``[%chess-incoming !>(~(tap by challenges-received))]
+  ::
+  ::  XX: should this be changed to [%y %games ~]?
+  ::      it lists all games, not just one
```
ashelkovnykov: Sure, this is probably for the best

src/urbit/app/chess.hoon
```diff
@@ -653,9 +674,22 @@
...
+  ::
+  ::  .^(noun %gx /=chess=/challenges/outgoing/noun)
+  ::  list challenges sent
+    [%x %challenges %outgoing ~]
+      ``[%chess-outgoing !>(~(tap by challenges-sent))]
+  ::
+  ::  .^(noun %gx /=chess=/challenges/incoming/noun)
+  ::  list challenges received
+    [%x %challenges %incoming ~]
+      ``[%chess-incoming !>(~(tap by challenges-received))]
```
ashelkovnykov: This shouldn't need two separates `mark`s. There should just be one `mark` called `%chess-challenges` which covers both cases, since the types are identical. Should suffice to delete one of `incoming.hoon` or `outgoing.hoon` and rename the remaining one to `challenges.hoon`.

ashelkovnykov (changes requested):
If an opponent declines a challenge, we need a way to remove the sent challenge from our frontend to reflect this. I think we'll need to add a `%challenge-declined` update, which we send to the frontend as a card along the "update" wire. When the frontend receives this, it'll remove the declined challenge from its set of sent challenges.

bonbud-macryg:
I've done some frontend work and we're most of the way there with this whole issue, but right now I'm a bit confused about wires.

I've added a `%challenge-declined` update to `update.hoon` but there's no one `/updates` wire to send to; there are only `<game-id>/updates` wires created once a challenge has been accepted and the game has been generated. I could create a general `/updates` wire, but I think you've mentioned wanting to split up the wires, scries, and marks into smaller jobs in future.

When ~tex either challenges ~mex, or declines a challenge from ~mex, ~tex sends a `%fact` to ~mex's `/poke/challenge` wire. This seems like the correct place from which to send a `%fact` to the frontend, which tells it to delete the challenge from the UI. From here we could either:
* Create a new wire like `/challenges-declined` to take this kind of card
* Do some conditional logic in `/challenges-sent` based on the "head" of the JSON being sent over. (e.g. "is the first string in this JSON `'challenge'` or `'challenge-declined'`?")

I tried the `/challenges-declined` idea and had a `declineChallenge` function implemented in earlier commits, but while I had everything compiling the outgoing challenges weren't being removed until I refreshed the page. (I've since removed that stuff and the behaviour is unchanged. I think when I refresh, the frontend is subscribing to the wires anew and getting the current state of `challenges-sent`. I understand the frontend better having tried this, but my `declineChallenge` stuff wasn't successfully doing anything!) Right now, doing that would also involve a `[%challenges-declined ~]  :_  this  [~]` in the agent's `on-watch`, which feels clumsy; there's nothing stored in a `challenges-declined` part of the state which we can send on the initial subscription.

ashelkovnykov:
@bonbud-macryg What `/poke/challenge` is specifying is the path on which your local `%chess` agent expects acks/nacks from the other user's agent. You can verify this by checking the `on-agent` arm of `%chess`: when an opposing agent nacks a challenge request, the code in our agent will remove the request from our set of challenges sent.

I think the most simple way to handle this is to have a single `/challenges` wire. On that wire, we send updates to the frontend:
1. `%challenge-sent`: When we send a challenge and it gets acked. This is also the point at which we should add the challenge to the `challenges-sent` set on the backend. Should be in the `on-agent` arm.
2. `%challenge-received`: When we receive a challenge from another user. Should be in the `on-poke` arm.
3. `%challenge-declined`: When an opponent accepts/declines our challenge. Sent from `on-poke` if the opponent declined our challenge, or `on-watch` if the opponent accepted our challenge.
4. `%challenge-resolved`: When we respond to an opponent's challenge and it gets acked. This is in `on-agent`, whether or not we're accepting or declining.

The idea is that updates about challenges sent/received will be used by the frontend to add challenges to its sets, and updates about challenges declined/resolved will be used to remove challenges from the sets. Right now, the frontend will manage the state of incoming challenges on its own, without receiving an update from the backend. This is incorrect behaviour which makes it possible for the frontend and backend states to diverge.

You'll note that my design above makes it a condition that messages we send first need to be acked. I think this is an important correctness check that `%chess` should be doing, and most of the time is not. However, performing these checks does result in some weird edge-cases. For example: what if an opponent sends a challenge to me and then uninstalls the chess app? Now, I can't accept/decline the challenge because his app no longer exists to ack my app. I think the solution to this is the option to "force delete" a challenge/game. However, that is a separate issue, and shouldn't be a concern here.

src/urbit/app/chess.hoon
```diff
@@ -706,11 +707,10 @@
...
-            ::  do nothing
-            ::[~ this]
+            ::  if opponent has acked our acceptance/rejection,
+            ::  confirm that to our frontend
             :-  :~  :*  %give  %fact   ~[/challenges]
-                        %chess-update  !>([%challenge-sent src.bowl])
+                        %chess-update  !>([%challenge-resolved src.bowl])
                     ==
                 ==
             this
```
bonbud-macryg, line 809: @ashelkovnykov Should the new `%challenge-sent` card go here, so that it's not being added to `challenges-sent` until the ack comes in from the ship we've challenged? If so, how will I access the `chess-challenge` info that needs to be sent to the frontend alongside the opponent ship's `@p`?

src/urbit/app/chess.hoon
```diff
@@ -5,6 +5,7 @@
 =,  chess
 ::
 ::  define state structures
+!:
```
ashelkovnykov: Need to remove before merge

src/urbit/app/chess.hoon
ashelkovnykov: That's a good point. What I'm thinking:
- Split the poke acks into two wires: `[%poke %challenge %send ~]` and `[%poke %challenge %resolve ~]`
- When we send a challenge in `on-poke`, add it to the set of outgoing challenges (looks like your current code still does). If the request is nacked, remove it from the set of outgoing challenges. If it the request is acked, lookup the `challenge` data by the `@p` of the opponent in the outgoing challenges set, and send that in the update card.

src/urbit/sur/chess.hoon
```diff
@@ -243,6 +243,7 @@
 ::  subscriber may receive as a %fact from the chess agent
 +$  chess-update
   $%  [%challenge who=ship challenge=chess-challenge]
+      [%challenge-declined who=ship]
```
ashelkovnykov: Need to change `%challenge` to `%challenge-sent`, as well as add `%challenge-received` and `%challenge-declined`

src/urbit/mar/chess/update.hoon
```diff
@@ -9,15 +9,41 @@
   |%
   ++  noun  upd
   ++  json
+    ::  XX: split %chess-update and /updates into several marks/wires
+    ::
+    ::      we shouldn't have all this information in one wire.
+    ::      as the chess app grows this will become unwieldy,
+    ::      subscribers to a wire only want the relevant info.
     ?-  -.upd
+      ::  XX: remove this?
```
ashelkovnykov: That's right - it should be safe to drop

src/frontend/tsx/App.tsx
```diff
@@ -2,13 +2,13 @@ import React from 'react'
...
 import useChessStore from '../ts/state/chessStore'
-import { GameInfo, ChallengeUpdate } from '../ts/types/urbitChess'
+import { GameInfo, Ship, ChallengeUpdate } from '../ts/types/urbitChess'
 import { Chessboard } from './Chessboard'
 import { Menu } from './Menu'
 import { GamePanel } from './GamePanel'
 
 export function App () {
-  const { urbit, setUrbit, receiveChallenge, receiveGame } = useChessStore()
+  const { urbit, setUrbit, receiveChallenge, sentChallenge, receiveGame } = useChessStore()
```
ashelkovnykov: Added some imports here, but don't seem to be using them

src/frontend/ts/types/urbitChess.ts
```diff
@@ -99,7 +104,12 @@ export interface ChessUpdate {
 }
 
 export interface ChallengeUpdate extends ChessUpdate {
-  chessUpdate: Update.Challenge
+  // XX: remove Challenge
+  chessUpdate: Update.Challenge |
+               Update.ChallengeSent |
+               Update.ChallengeReceived |
+               Update.ChallengeDeclined |
+               Update.ChallengeResolved
```
ashelkovnykov: Not quite - you're going to want something like:
```
export interface ChallengeUpdate extends ChessUpdate {
  chessUpdate: Update.ChallengeSent |
               Update.ChallengeReceived |
               Update.ChallengeDeclined |
               Update.ChallengeResolved
  who: Ship
}

export interface ChallengeSentUpdate extends ChallengeUpdate {
  challengerSide: Side
  event: string
  round: string
}

export interface ChallengeReceivedUpdate extends ChallengeUpdate {
  challengerSide: Side
  event: string
  round: string
}

export interface ChallengeDeclinedUpdate extends ChallengeUpdate {}

export interface ChallengeResolvedUpdate extends ChallengeUpdate {}
```

src/frontend/ts/types/urbitChess.ts
```diff
@@ -28,7 +28,12 @@ export enum Result {
 }
 
 export enum Update {
+  // XX: remove Challenge?
```
ashelkovnykov: Yeah, safe to remove

src/frontend/ts/state/chessStore.ts
```diff
@@ -10,11 +10,43 @@ const useChessStore = create<ChessState>((set, get) => ({
...
+        case Update.ChallengeReceived: {
+          set(state => ({ incomingChallenges: state.incomingChallenges.set(data.who, data as Challenge) }))
+          break
+        }
+        case Update.ChallengeDeclined: {
+          get().outgoingChallenges.delete(data.who)
+          break
+        }
+        case Update.ChallengeResolved: {
+          // XX: what should this do?
```
ashelkovnykov: `get().incomingChallenges.delete(data.who)`

src/frontend/ts/state/chessStore.ts
```diff
@@ -10,11 +10,43 @@ const useChessStore = create<ChessState>((set, get) => ({
...
+          console.log((data as ChallengeUpdate).chessUpdate)
+          console.log((data as ChallengeUpdate).who)
+          console.log((data as ChallengeUpdate).challengerSide)
+          console.log((data as ChallengeUpdate).event)
+          console.log((data as ChallengeUpdate).round)
+        }
+      }
+    },
+  sentChallenge: (data: ChallengeUpdate) =>
+    set(state => ({ outgoingChallenges: state.outgoingChallenges.set(data.who, data as Challenge) })),
```
ashelkovnykov: Not needed

src/frontend/ts/state/chessState.ts
```diff
@@ -7,10 +7,12 @@ interface ChessState {
   practiceBoard: String | null;
   activeGames: Map<GameID, ActiveGameInfo>;
   incomingChallenges: Map<Ship, Challenge>;
+  outgoingChallenges: Map<Ship, Challenge>;
   setUrbit: (urbit: Urbit) => void;
   setDisplayGame: (displayGame: ActiveGameInfo | null) => void;
   setPracticeBoard: (practiceBoard: String | null) => void;
   receiveChallenge: (data: ChallengeUpdate) => void;
+  sentChallenge: (data: ChallengeUpdate) => void;
```
ashelkovnykov: Not needed

ashelkovnykov (changes requested):
Very close to done

src/urbit/app/chess.hoon
bonbud-macryg: Sounds good, I was wondering if the solution was something like that second point.

src/urbit/app/chess.hoon
```diff
@@ -749,25 +770,57 @@
...
+                ==
+            this
+          ::  if not, print the error and
+          ::  consider the challenge declined
           %-  (slog u.p.sign)
-          :-  ~
+          :-  :~  :*  %give  %fact   ~[/challenges]
+                      %chess-update  !>([%challenge-declined src.bowl])
+                  ==
+              ==
```
ashelkovnykov: Should we consider it declined? I think it's fine for now, though later it should be one of those cases where we send a notification to the user that an error occurred (the challenge never arrived).

src/frontend/tsx/App.tsx
```diff
@@ -2,7 +2,7 @@ import React from 'react'
 import { Beforeunload } from 'react-beforeunload'
 import Urbit from '@urbit/http-api'
 import useChessStore from '../ts/state/chessStore'
-import { GameInfo, ChallengeUpdate } from '../ts/types/urbitChess'
+import { GameInfo, Ship, ChallengeUpdate } from '../ts/types/urbitChess'
```
ashelkovnykov: Any reason for this import?

src/frontend/ts/state/chessStore.ts
```diff
@@ -10,11 +10,37 @@ const useChessStore = create<ChessState>((set, get) => ({
...
   receiveChallenge: (data: ChallengeUpdate) =>
-    set(state => ({ incomingChallenges: state.incomingChallenges.set(data.who, data as Challenge) })),
+    {
+      switch (data.chessUpdate) {
+        case Update.ChallengeSent: {
+          set(state => ({ outgoingChallenges: state.outgoingChallenges.set(data.who, data as ChallengeSentUpdate) }))
+          break
+        }
+        case Update.ChallengeReceived: {
+          set(state => ({ outgoingChallenges: state.outgoingChallenges.set(data.who, data as ChallengeReceivedUpdate) }))
```
ashelkovnykov: Bug here: forgot to replace `outgoingChallenges` with `incomingChallenges` after copy

src/frontend/ts/state/chessStore.ts
```diff
@@ -44,6 +70,8 @@ const useChessStore = create<ChessState>((set, get) => ({
       }
     }
 
+    // XX: add new Update.Foo for each new branch
+    //     in update.hoon
```
ashelkovnykov: No, that's not necessary. Any other update type should be handled by the default case (log bad update). Besides, this is really an issue with all updates sharing a single type.

ashelkovnykov (changes requested):
Almost there. Three things:
1. See minor comments below
2. Need to add the incoming/outgoing challenge sets to one of the existing React hooks, or a new one. Currently, declining/accepting a challenge doesn't update either agent's frontend correctly without refreshing.
3. Need to make the "incoming"/"outgoing" spans clickable/buttons/tabs, or need to add some sort of visual differentiation between the incoming/outgoing challenges (they're just concatenated right now, though the cards do look different)

Other than that, it's functioning correctly.

src/urbit/app/chess.hoon
bonbud-macryg: Yeah maybe we should explicitly tell the player about nacks, either printing to terminal or on the frontend. I've added a todo about it.

bonbud-macryg:
@ashelkovnykov One more issue to address on this PR and it's pretty much ready. When you send a challenge the event description is either 'black', 'white', or 'random' depending on the side you picked.



When you refresh the page, it then shows the correct description.



It's late here so I'll have to look into it tomorrow. Any initial suspicions as to what might be causing that?

src/frontend/ts/state/chessStore.ts
```diff
@@ -34,6 +60,15 @@ const useChessStore = create<ChessState>((set, get) => ({
...
     })
+
+    const opp: string = data.white === window.ship ? data.black : data.white 
+    if (get().outgoingChallenges.has(opp)) {
+      if (data.event === get().outgoingChallenges.get(opp).event) {
+        // XX: should this take into account our and our opponent's
+        //     sides to narrow it down further?
+        get().removeOutgoingChallenge(opp)
+      }
+    }
```
bonbud-macryg: This removes outgoing challenges when they're accepted by our opponent. I don't love it, but the first alternative that comes to mind is giving every challenge its own ID, which seems like overkill.

bonbud-macryg: Another option: change default `event` from 'random', 'black', or 'white' to something like '~2022.12.7'. This would avoid collision of default event names, but still feels a bit clumsy.

bonbud-macryg: The more I think about this second option the more sense it's making to me. We've talked about a "quick challenge" button for friends and in that case the generated event description that gets created for it *would* be something like `~2022.12.7`.

This also limits the issue to the frontend. Which is either a plus or a minus depending on if one thinks `challenge-id`s should be part of the "core" Chess implementation.

bonbud-macryg:
On the outgoing challenge descriptions, we could also just get rid of them entirely ([image](https://0x0.st/odSC.png)). You can only have one outgoing challenge to one ship at a time, so there's no need to distinguish between sets of them.

This doesn't solve our updating issue as the piece icons have the same problem.

bonbud-macryg:
I think I've narrowed this bug down to the backend. When `receiveGame()` gets a `challengeSent` update upon refreshing the page, everything is fine. When `receiveGame()` gets that update when we send a challenge, it breaks. Compare the `~&` output from these two cases.

When we refresh the page:
```
>>  ~bud
>>  %random
>>  '12.8 3'
>>  [~ ~]
>>  ~dev
>>  %black
>>  '12.8 2'
>>  [~ ~]
>>  ~ser
>>  %random
>>  '12.8 6'
>>  [~ ~]
>>  ~bud
>>  %random
>>  '12.8 3'
>>  [~ ~]
>>  ~dev
>>  %black
>>  '12.8 2'
>>  [~ ~]
>>  ~ser
>>  %random
>>  '12.8 6'
>>  [~ ~]
```

When we send a new challenge:
```
>>  ~ser
ames: ~ser ack crashed %exit
>>  ~ser
>>  ~ser
ames: ~ser ack crashed %exit
>>  ~ser
>>  ~ser
ames: ~ser ack crashed %exit
>>  ~ser
; ~ser not responding still trying
```

`challenger-side`, `round`, and `event` information aren't making it to `update.hoon`. There are also those `ames` errors which get printed intermittently after the `%challenge` poke is sent. I don't think this is related to the bug, it happens on some `%challenge` pokes and not others.

bonbud-macryg:
Narrowing down further, I think I've got it.

***

* ~nec sends challenges to ~ser
* At the moment of `sendChallenge()`, `who`, `side`, and `description` are all correct in ~nec’s browser
* By the time they reach `%challenge-sent` in ~nec’s `update.hoon`, only `who` remains.
	* What’s different?
* ~ser actually receives the challenge correctly
```
>>  “received who.upd: ~nec”
>>  “received challenger.upd: %white”
>>  “received event.upd: ’12.8 3’”
>>  “received round.upd: [~ ~]”
```
* So why don’t they make it to ~nec’s `%challenge-sent` to itself?

By hard-coding the following on the `/poke/challenge` wire…
```
:-  :~  :*  %give  %fact   ~[/challenges]
            %chess-update  !>([%challenge-sent src.bowl [challenger-side=%random event='ON-AGENT' round=`~]])
        ==
    ==
```

…we get the following in ~nec’s dojo upon running `sendChallenge()`…
```
>   "sent who.upd: ~ser"
>   "sent challenger.upd: %random"
>   "sent event.upd: 'ON-AGENT'"
>   "sent round.upd: [~ ~]"
```

…~ser receives the challenge with the “real” event description. The `/poke/challenge` wire is just for ~nec’s benefit, and it’s not sending the right information.

What is *is* sending maps exactly onto the default values described in `/sur/chess.hoon`. Therefore, I suspect that the existing card in the `/poke/challenge` wire, below, is sending nothing but `src.bowl`, and then at some point the rest of the info is filled in with default values.
```
:-  :~  :*  %give  %fact   ~[/challenges]
            %chess-update  !>([%challenge-sent src.bowl (~(get by challenges-sent) src.bowl)])
        ==
    ==
```

src/urbit/app/chess.hoon
```diff
@@ -90,6 +90,9 @@
               "already challenged {<who.action>}"
             :~  [%give %poke-ack `~[leaf+err]]
             ==
+          ::  XX: add error on attempting to send a
+          ::      challenge to yourself
+          ::
```
ashelkovnykov, line 93: That's actually not necessary - I often check changes in my fake ships by sending a challenge to myself and playing both sides in one interface

src/urbit/app/chess.hoon
```diff
@@ -786,13 +789,17 @@
     [%poke %challenge %send ~]
       ?+  -.sign  (on-agent:default wire sign)
         %poke-ack
-          ::  if opponent has acked our challenge, do nothing
+          ::  if opponent has acked our challenge, add it to the frontned
           ?~  p.sign
+          ::  
             :-  :~  :*  %give  %fact   ~[/challenges]
-                        %chess-update  !>([%challenge-sent src.bowl (~(get by challenges-sent) src.bowl)])
+                        %chess-update  !>([%challenge-sent src.bowl q:(rear ~(tap by challenges-sent))])
```
ashelkovnykov: Left a comment before your PR a few minutes ago, but editing it now for your latest commit: this should just be `got`

src/frontend/ts/state/chessStore.ts
```diff
@@ -186,6 +193,12 @@ const useChessStore = create<ChessState>((set, get) => ({
 
     set({ incomingChallenges })
   },
+  removeOutgoingChallenge: (who: Ship) => {
+    let outgoingChallenges: Map<Ship, Challenge> = get().outgoingChallenges
+    outgoingChallenges.delete(who)
+
+    set({ outgoingChallenges })
+  },
```
ashelkovnykov: Sorry, I should have left a comment about this earlier. The `removeOutgoing` function in `ChessState` that's used in `Challenges` is actually precisely the thing that we're trying to get rid of (as well as the `onSuccess` callbacks in `acceptChallenge` and `declineChallenge`). This should still work as you have written by copying the contents of `removeChallenge` and `removeOutgoingChallenge` into the `receiveChallenge` function (which probably should be renamed to `receiveChallengeUpdate`):
```
...
case Update.ChallengeDeclined: {
  let outgoingChallenges: Map<Ship, Challenge> = get().outgoingChallenges
  outgoingChallenges.delete(data.who)

  set({ outgoingChallenges })
  break
}
case Update.ChallengeResolved: {
  let incomingChallenges: Map<Ship, Challenge> = get().incomingChallenges
  incomingChallenges.delete(data.who)

  set({ incomingChallenges })
  break
}
...
```

src/frontend/ts/state/chessState.ts
```diff
@@ -15,6 +15,7 @@ interface ChessState {
   receiveGame: (data: GameInfo) => void;
   receiveUpdate: (data: ChessUpdate) => void;
   removeChallenge: (who: Ship) => void;
+  removeOutgoingChallenge: (who: Ship) => void;
```
ashelkovnykov: See above comment - should be removed

ashelkovnykov (changes requested):
Need to take a look at one more thing - might be another comment in a moment

src/urbit/app/chess.hoon
```diff
@@ -749,25 +773,61 @@
...
             challenges-sent  (~(del by challenges-sent) src.bowl)
           ==
       ==
     ::
+    :: get ack/nack when we accept/reject a challenge
+    [%poke %challenge %resolve ~]
+      ?+  -.sign  (on-agent:default wire sign)
+        %poke-ack
+          ?~  p.sign
+          ::  if opponent has acked our acceptance/rejection,
```
ashelkovnykov: This case is only for when our opponent acks a rejection, not acceptance. We're forgetting to send `%challenge-resolved` cards when the opponent accepts our game (i.e. subscribes to us): after lines `664` and `698` in the code for `[%game @ta %moves ~]` in `on-watch`.

src/frontend/ts/state/chessStore.ts
```diff
@@ -14,33 +14,45 @@ const useChessStore = create<ChessState>((set, get) => ({
...
+    switch (data.chessUpdate) {
+      case Update.ChallengeSent: {
+        set(state => ({ outgoingChallenges: state.outgoingChallenges.set(data.who, data as ChallengeSentUpdate) }))
+        break
+      }
+      case Update.ChallengeReceived: {
+        set(state => ({ incomingChallenges: state.incomingChallenges.set(data.who, data as ChallengeReceivedUpdate) }))
+        break
+      }
+      case Update.ChallengeDeclined: {
```
ashelkovnykov: See primary comment for this review. This should be renamed to `ChallengeResolved`.

src/frontend/ts/state/chessStore.ts
```diff
@@ -14,33 +14,45 @@ const useChessStore = create<ChessState>((set, get) => ({
...
+        let outgoingChallenges: Map<Ship, Challenge> = get().outgoingChallenges
+        outgoingChallenges.delete(data.who)
+
+        set({ outgoingChallenges })
+        break
+      }
+      case Update.ChallengeResolved: {
```
ashelkovnykov: See primary comment for this review. This should be renamed to `ChallengeReplied` and it should only remove entries from `incomingChallenges`.

src/urbit/app/chess.hoon
```diff
@@ -154,6 +152,10 @@
                   :*  %give  %fact  ~[/active-games]
                       %chess-game  !>(new-game)
                   ==
+                  ::  tell our frontend we accepted the challenge
+                  :*  %give  %fact   ~[/challenges]
+                      %chess-update  !>([%challenge-resolved who.action])
```
ashelkovnykov: See primary comment for this review. This should now be `%challenge-replied`.

src/urbit/app/chess.hoon
```diff
@@ -662,6 +664,10 @@
                 :*  %give  %fact  ~[/active-games]
```
ashelkovnykov, line 664: See primary comment for this review. Line `421` should now be `%challenge-resolved`.

src/urbit/app/chess.hoon
```diff
@@ -793,7 +803,7 @@
           ?~  p.sign
           ::  
             :-  :~  :*  %give  %fact   ~[/challenges]
-                        %chess-update  !>([%challenge-sent src.bowl u.+:(~(get by challenges-sent) src.bowl)])
+                        %chess-update  !>([%challenge-sent src.bowl (~(got by challenges-sent) src.bowl)])
                     ==
                 ==
             this
```
ashelkovnykov, line 809: See primary comment for this review. Line `833` should now be `%challenge-replied`, and the wire on lines `180`/`826` should now be `/poke/challenge/reply`/`[%poke %challenge %reply ~]`.

src/urbit/app/chess.hoon
```diff
@@ -812,18 +822,19 @@
           ==
```
ashelkovnykov, line 819: The card on lines `816` - `819` should be removed. If a sent challenge gets nacked, we don't need to send an update card to the frontend, since we have nothing to remove from the frontend state, since we never sent a card to add this challenge to the frontend.

ashelkovnykov (changes requested):
Sorry - I was unclear in my original write-up. We should only have 4 challenge updates, which directly map to adding/removing incoming/outgoing challenges. So:
- `challengeSent` -> add to outgoing
- `challengeReceived` -> add to incoming
- `challengeReplied` -> remove from incoming
- `challengeResolved` -> remove from outgoing

Aside from this, everything looks correct. After these quick renamings and squashing the commits, we should be good to merge.

src/frontend/ts/types/urbitChess.ts
```diff
@@ -104,13 +107,29 @@ export interface ChessUpdate {
...
+  round: string
+}
+
+export interface ChallengeReceivedUpdate extends ChallengeUpdate {
   challengerSide: Side
   event: string
   round: string
 }
 
+export interface ChallengeDeclinedUpdate extends ChallengeUpdate {}
```
ashelkovnykov: I originally wrote a comment about how this object needs to be renamed, but it looks like we don't even use it, so we can just delete it and `ChallengeResolvedUpdate` too.

src/frontend/ts/types/urbitChess.ts
```diff
@@ -104,13 +107,29 @@ export interface ChessUpdate {
...
   who: Ship
+}
+
+export interface ChallengeSentUpdate extends ChallengeUpdate {
+  challengerSide: Side
+  event: string
+  round: string
+}
+
+export interface ChallengeReceivedUpdate extends ChallengeUpdate {
```
ashelkovnykov: Each of these two should have the specific enum in the `chessUpdate` field. e.g.:
```
export interface ChallengeSentUpdate extends ChallengeUpdate {
  chessUpdate: Update.ChallengeSent
  challengerSide: Side
  event: string
  round: string
}
```

ashelkovnykov (approved):
LGTM. Two minor comments, and linter errors. Confirmed everything working as intended on two fake ships.

Issue #14, opened 2022-07-21
`%pals` integration


ashelkovnykov:
Quick-select feature for challenging your friends.

Frontend and backend work.

PR #46, opened 2022-09-25
Add %friends scry path and %chess-pals mark

bonbud-macryg:
Rough-draft backend implementation for #14. I can't get a working scry to the `%mutuals` path and it's not that I'm pointing at the wrong location, so I can't test whether this works or not.

This currently uses paldev's `/lib/pals.hoon` library, as recommended in the comments of the `%pals` app. I thought I'd do it the way he recommends as he's not going to break that in future updates, but it might be needlessly complicated. (I also thought this might be very easily extensible. If a future dev wants to scry their own `%leeches` path, they're just copy-pasting the `%mutuals` scry.)

An alternative approach would be to scry `%pals` directly with a `.^` rune in `app/chess.hoon`. I've performed working scries with that, but not with this `%chess-partners` mark. I don't quite get how the two marks in the `.^` expression interact with eachother.

I had a look at other apps to see how they interface with `%pals`. `%rumors` uses the library I use here, but `%escape` seems to do it all on the frontend.


ashelkovnykov:
Did you mean that this is a backend implementation for #14? #7 could be done with a scry, though I'm not sure whether that's possible/correct. The easiest solution for #7 is to add another subscribe path to `on-watch` and an associated subscription from the frontend for outgoing challenges.

Oh man, the syncing of state between the frontend and backend is so inconsistent... The code needs a deep clean.

bonbud-macryg:
Oops, yes this is #14.

src/urbit/app/chess.hoon
```diff
@@ -570,6 +573,10 @@
         ?~  archived-game  ~
         ``[%chess-game !>(u.archived-game)]
       ``[%chess-game !>(game.u.active-game)]
+  ::
+    [%x %pals %mutuals ~]
+      ``[%chess-partners !>(~(tap in (mutuals:pals ~.)))]
```
ashelkovnykov: 1. I think "friends" or "pals" would be better than "partners"
2. I believe that you need to pass the bowl into this to make it work
3. Leave it as a `(set @p)` and let the `mark` deal with it
4. The scry path can be simplified; we can hide some of the implementation details. The only real options IMO are "mutuals" or "targets". Maybe in the future limit it to people with the "chess" tag?

Final product:
```
[%x %friends ~]
  ``[%chess-friends !>((~(mutuals pals bowl) ~))
```

src/urbit/mar/chess/partners.hoon
```diff
@@ -0,0 +1,19 @@
+=,  format
+|_  partners=`(list @p)`update
```
ashelkovnykov: Building on the previous comment, this should be:
```
|_  friends=(set @p)
```

src/urbit/mar/chess/partners.hoon
```diff
@@ -0,0 +1,19 @@
...
+++  grab
+  |%
+  ++  noun  `(list @p)`update
```
ashelkovnykov: Think you got confused, should just be `++  noun  friends` (after changes from previous comments)

src/urbit/mar/chess/partners.hoon
```diff
@@ -0,0 +1,19 @@
...
+++  grow
+  |%
+  ++  noun  partners
+  ++  json
+    %-  frond:enjs
+    :-  'mutuals'
+    :-  %a
+    %-  turn
+      partners
+    |=(part=@p (ship:enjs part))
```
ashelkovnykov:
```
%+  turn  friends
|=  fren=@p  (ship:enjs fren)
```

ashelkovnykov (changes requested):
`.^` is a virtual rune that only exists in virtual Hoon - which is what is used in Gall and Dojo. I'll let you play with it yourself, but you can have your agents do scries inside scries/pokes and it Just Works.

I think doing a scry to `%chess` and it doing its own scry to `%pals` is the correct way to go. It leaves us free to change how we do friends/contacts later. It also keeps `%chess` working if the user doesn't have `%pals`: we can handle a scry failure and return an empty list. We can then omit rendering the "quick-challenge friends" tab if the result of the scry is an empty list. Note that this also works for if the user has `%pals` but does not have any pals.

src/urbit/mar/chess/pals.hoon
```diff
@@ -0,0 +1,18 @@
...
+  --
+++  grow
+  |%
+  ++  noun  pals
+  ++  json
+    %-  frond:enjs
+    :-  'mutuals'
+    :-  %a
+    %-  turn  pals
+    |=(fren=@p (ship:enjs fren))
```
bonbud-macryg: Just realised after pushing this change I'm trying to `turn` a set, which will fail. Will fix.

src/urbit/app/chess.hoon
```diff
@@ -653,9 +654,15 @@
...
-    ::  collect all the game-id keys
+  ::
+  ::  .^(noun %gx /=chess=/mutuals/noun)
+  ::  .^(json %gx /=chess=/mutuals/json)
+  ::  read mutual pals
+    [%x %mutuals ~]
+      ``[%chess-pals !>((~(mutuals pals bowl) ~.))]
+  ::
+  ::  .^(arch %gy /=chess=/game)
+  ::  collect all the game-id keys
```
bonbud-macryg: This wasn't added in this commit, just moved over to the left. I've amended the indentation of the comments here to match the style we see in `/sur/chess`'s definition of `chess-game` and `chess-position`. If we didn't have the comments here, I'd keep the single `::` runes separating the scry paths, which is why I've indented the comments this way.

src/urbit/app/chess.hoon
```diff
@@ -653,9 +654,15 @@
...
       ``[%chess-game !>(game.u.active-game)]
-    ::
-    ::  .^(arch %gy /=chess=/game)
-    ::  collect all the game-id keys
+  ::
+  ::  .^(noun %gx /=chess=/mutuals/noun)
+  ::  .^(json %gx /=chess=/mutuals/json)
+  ::  read mutual pals
+    [%x %mutuals ~]
+      ``[%chess-pals !>((~(mutuals pals bowl) ~.))]
```
bonbud-macryg: I'll mention here that the mark is called `chess-pals` rather than `chess-mutuals` because we could, and maybe will, extend that same mark to deal with different calls like `chess-update` does.

src/urbit/lib/pals.hoon
```diff
@@ -0,0 +1,18 @@
...
+++  mutual   |=  [list=@ta =ship]  (s _| /mutuals/[list]/(scot %p ship))
+::
+++  base     ~+  /(scot %p our)/pals/(scot %da now)
+++  running  ~+  .^(? %gu base)
+::
+++  s
+  |*  [=mold =path]  ~+
+  ?.  running  *mold
+  .^(mold %gx (weld base (snoc `^path`path %noun)))
+--
```
bonbud-macryg: This is palfun's `pals` library, taken from `%rumors` and unedited.

rovmug-ticfyn (commented):
LGTM

bonbud-macryg:
Pushed one last amendment, just adding a newline at the end of `/lib/pals.hoon`.

src/urbit/app/chess.hoon
```diff
@@ -1,7 +1,8 @@
 ::  chess: fully decentralized, peer-to-peer chess app for urbit
 ::
 ::  import libraries and expose namespace
-/+  chess, dbug, default-agent
+
```
ashelkovnykov: Unnecessary white space

src/urbit/app/chess.hoon
```diff
@@ -641,9 +642,9 @@
   |=  =path
   ^-  (unit (unit cage))
   ?+  path  (on-peek:default path)
-    ::
-    ::  read game info
-    ::  either active or archived
+  ::
+  ::  read game info
+  ::  either active or archived
```
ashelkovnykov: I think the indentation for the comments was better as before: indented to the same depth as the code which it is annotating

src/urbit/app/chess.hoon
ashelkovnykov, line 737: Should we be targeting mutual contacts, or any contact we've added to `%pals`? "Targets" might be the better wire, since that's the set of everyone we like enough to send them a friend request (whether or not they responded).

src/urbit/mar/chess/pals.hoon
```diff
@@ -0,0 +1,18 @@
...
+++  grab
+  |%
+  ++  noun  pals
+  --
+++  grow
+  |%
+  ++  noun  pals
+  ++  json
+    %-  frond:enjs
+    :-  'mutuals'
```
ashelkovnykov: While `%pals` uses the term "mutuals", I don't think that we should:
1. It might be that we should be querying "targets" instead of "mutuals" (see above comment)
2. "Mutuals" doesn't mean anything to `%chess`, but it does to `%pals`. By returning `mutuals` as a mark for the scry, we're leaking information about what we use on the backend, and forwarding it to a user who may have no clue what the context is.
3. The mark will need to be changed, or else make no sense, if we ever change the backend (for example, to `%whom` instead of `%pals` - or even to the union of both)

We should go with something much more generic, like "friends" or "contacts"

ashelkovnykov (changes requested):

src/urbit/app/chess.hoon
bonbud-macryg: Resolved. There should also be an example scry for `[%x %game @ta ~]` but I think that's added in #45.

src/urbit/mar/chess/pals.hoon
bonbud-macryg: The idea was to keep things sort of modular and extensible, maybe in future someone wants to add more wires for pals/contacts to be handled in `%chess-pals`, but this makes sense. Let's go with `:-  'friends'`.

src/urbit/app/chess.hoon
bonbud-macryg: Either works, we could support both in some capacity, but I chose mutuals because:

1. Cozier.
2. I wonder if `%pals`-based functionality should lean more towards mutuals than "targets". I've been `target`ed by randos I've never talked to and never see, and I don't want them seeing my mutuals' rumors, for example. `target`ing someone just to access `%pals`-based functionality is kind of spammy and annoying, it shouldn't be encouraged. Something like Facebook gave you very granular permissions about who could see/do what based on whether or not you were Friends with them, and you wouldn't want them extending those permissions to every random person who'd sent you a friend request.
3. To put myself in the randos' position: just because I like someone enough to add them, doesn't mean they like me back :( Are they ignoring my outgoing challenge that's sat there for a few days? Am I being annoying? It's partly why mutuals is cozier.
4. So a game between mutuals *feels* different. Could it work differently? We've talked about the "quick challenge" button, what else might people want to add in future? Example: maybe if our pal breaches in the middle of a long correspondence game, their `%chess` app pokes mutuals for unfinished active games and we can pick up where we left off. And maybe people will want to add more of these privileges to these "friendly games" between people they know and trust. Who knows; a game between mutuals feels like a meaningfully different thing, and something that could be built on top of.

src/urbit/mar/chess/pals.hoon
ashelkovnykov: Your explanation for your reasoning makes sense. My gut feeling is that it's better for us to define our own API and terminology, just so that we're not subject to changes in the app availability/design landscape (e.g. what apps will become popular and how they're written are out of our control, so let's just introduce our own idea of "friends" and we can adjust what that means on the backend as Urbit evolves).

src/urbit/app/chess.hoon
ashelkovnykov: I think that's a very solid rebuttal. Let's stick to mutuals.

src/urbit/app/chess.hoon
```diff
@@ -653,9 +654,15 @@
...
         ``[%chess-game !>(u.archived-game)]
       ``[%chess-game !>(game.u.active-game)]
-    ::
-    ::  .^(arch %gy /=chess=/game)
-    ::  collect all the game-id keys
+  ::
+  ::  .^(noun %gx /=chess=/mutuals/noun)
+  ::  .^(json %gx /=chess=/mutuals/json)
+  ::  read mutual pals
+    [%x %mutuals ~]
```
ashelkovnykov: Though we're targeting the `%mutuals` scry path, we should still be calling it `%pals` or `%friends` on our end

src/urbit/app/chess.hoon
ashelkovnykov: What do ya know, you're right: this is the correct commenting style for kingside tall format (which we're also doing incorrectly). Okay, fair enough. Feel free to correct the style on any code you touch, and we'll do a style refactor after milestone 1, when we have a set style guide.

ashelkovnykov (commented):
You resolved a couple of the comments, but didn't commit the changes for them yet.

Aside from that, the backend is ready to be connected to the frontend.

ashelkovnykov (commented):
LGTM - did you want to include the frontend work in this PR?

bonbud-macryg:
I'm happy to merge this now if you are.

ashelkovnykov (approved):
Looks like some of the new changes conflict with this PR. Could you rebase or merge with the new master branch? Then we'll merge.

PR #47, opened 2022-09-28
Implement request to undo moves for #20

### LINKED ISSUE #20: Take back/undo moves
Allow players to request a take-back. Will require pulling popping old game states off the stack, and a new button + pop-up.

Frontend and backend work.

rovmug-ticfyn:
This partially resolves #20, only backend at this point

### DIFF
--- a/src/urbit/app/chess.hoon
+++ b/src/urbit/app/chess.hoon
@@ -19,6 +19,8 @@
       sent-draw-offer=?
       got-draw-offer=?
       auto-claim-special-draws=?
+      sent-undo-request=?
+      got-undo-request=?
   ==
 +$  state-1
   $:  %1
@@ -166,7 +168,7 @@
                                (~(del by challenges-sent) our.bowl)
                              challenges-sent
             ::  put our new game into the map of games
-            games  (~(put by games) game-id [new-game *chess-position *(map @t @ud) | | | | |])
+            games  (~(put by games) game-id [new-game *chess-position *(map @t @ud) | | | | | | |])
           ==
         %decline-game
           =/  challenge  (~(get by challenges-received) who.action)
@@ -432,6 +434,98 @@
             archive  (~(put by archive) game-id.action game.u.game-state(result `result))
           ==
       ==
+        %request-undo
+          =/  game  (~(get by games) game-id.action)
+          ?~  game
+            :_  this
+            =/  err
+              "no active game with id {<game-id.action>}"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          ::  send undo request to opponent
+          :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
+                      %chess-undo-request  !>(~)
+                  ==
+                  :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
+                      %chess-update  !>([%undo-request game-id.action])
+                  ==
+              ==
+          ::  record that undo has been requested
+          %=  this
+            games  (~(put by games) game-id.action u.game(sent-undo-request &))
+          ==
+        %decline-undo
+          =/  game  (~(get by games) game-id.action)
+          ?~  game
+            :_  this
+            =/  err
+              "no active game with id {<game-id.action>}"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          ::  check for open undo request
+          ?.  got-undo-request.u.game
+            :_  this
+            =/  err
+              "no undo request to decline for game {<game-id.action>}"
+              :~  [%give %poke-ack `~[leaf+err]]
+              ==
+          ::  decline undo request
+          :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
+                      %chess-undo-decline  !>(~)
+                  ==
+                  :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
+                      %chess-update  !>([%undo-declined game-id.action])
+                  ==
+              ==
+          %=  this
+            ::  record that undo request is gone
+            games  (~(put by games) game-id.action u.game(got-undo-request |))
+          ==
+        %accept-undo
+          =/  game-state  (~(get by games) game-id.action)
+          ::  check for valid game
+          ?~  game-state
+            :_  this
+            =/  err
+              "no active game with id {<game-id.action>}"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
+          =,  u.game-state
+          ::  check for open undo request
+          ?.  got-undo-request
+            :_  this
+            =/  err
+              "no undo request to decline for game {<game-id.action>}"
+              :~  [%give %poke-ack `~[leaf+err]]
+              ==
+          ::  tell opponent we accept the undo request
+          :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
+                      %chess-undo-accept  !>(~)
+                  ==
+                  ::  update observers that we accept the undo request
+                  :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
+                      %chess-update  !>([%undo-accepted game-id.action])
+                  ==
+              ==
+          =/  ship-to-move
+            ?-  player-to-move.position.u.game-state
+              %white  white.game
+              %black  black.game
+            ==
+          ?>  ?=([%ship @p] ship-to-move)
+          =:
+            moves.game.u.game-state  ?:  =(+.ship-to-move our.bowl)
+                                       (snip moves.game)
+                                     (snip (snip moves.game))
+            position.u.game-state  ?:  =(+.ship-to-move our.bowl)
+                                     (fen-to-position (tail (rear (snip moves.game))))
+                                   (fen-to-position (tail (rear (snip (snip moves.game)))))
+            got-undo-request.u.game-state  |
+          ==
+          %=  this
+            games  (~(put by games) game-id.action u.game-state)
+          ==
+      ==
     ::
     ::  handle incoming challenges
     %chess-challenge
@@ -529,7 +623,7 @@
                                  (~(del by challenges-sent) our.bowl)
                                challenges-sent
               rng-state  (~(del by rng-state) src.bowl)
-              games  (~(put by games) game-id [new-game *chess-position *(map @t @ud) | | | | |])
+              games  (~(put by games) game-id [new-game *chess-position *(map @t @ud) | | | | | | |])
             ==
           ::  we're the challenger
           ?>  ?=(^ her-hash.u.commitment)
@@ -568,7 +662,7 @@
               ==
           ==
       %=  this
-        games  (~(put by games) game-id.action [game.action u.new-position *(map @t @ud) (check-50-move-rule u.new-position) & | | |])
+        games  (~(put by games) game-id.action [game.action u.new-position *(map @t @ud) (check-50-move-rule u.new-position) & | | | | |])
       ==
     ::
     ::  directly inject game subscriptions (for debugging)
@@ -702,7 +796,7 @@
         %=  this
           challenges-sent  (~(del by challenges-sent) src.bowl)
           rng-state  (~(del by rng-state) src.bowl)
-          games  (~(put by games) u.game-id [new-game *chess-position *(map @t @ud) | & | | |])
+          games  (~(put by games) u.game-id [new-game *chess-position *(map @t @ud) | & | | | | |])
         ==
       ::  assign white and black to players if challenger chose
       =+  ^=  [white-player black-player]
@@ -739,7 +833,7 @@
           ==
       %=  this
         challenges-sent  (~(del by challenges-sent) src.bowl)
-        games  (~(put by games) u.game-id [new-game *chess-position *(map @t @ud) | & | | |])
+        games  (~(put by games) u.game-id [new-game *chess-position *(map @t @ud) | & | | | | |])
       ==
     ::
     ::  XX: document this
@@ -831,7 +925,7 @@
         %poke-ack
           ::  if opponent has acked our challenge, add it to the frontned
           ?~  p.sign
-          ::  
+          ::
             :-  :~  :*  %give  %fact   ~[/challenges]
                         %chess-update  !>([%challenge-sent src.bowl (~(got by challenges-sent) src.bowl)])
                     ==
@@ -1000,6 +1094,42 @@
                       ~
                   ==
               ==
+            %chess-undo-request
+              :-  :~  :*  %give  %fact  ~[/game/(scot %da u.game-id)/updates]
+                          %chess-update  !>([%undo-request u.game-id])
+                      ==
+                  ==
+              %=  this
+                games  (~(put by games) u.game-id u.game-state(got-undo-request &))
+              ==
+            %chess-undo-decline
+              :-  :~  :*  %give  %fact  ~[/game/(scot %da u.game-id)/updates]
+                          %chess-update  !>([%undo-declined u.game-id])
+                      ==
+                  ==
+              %=  this
+                games    (~(put by games) u.game-id u.game-state(sent-undo-request |))
+              ==
+            %chess-undo-accept
+              ?.  sent-undo-request.u.game-state
+                [~ this]
+              :-  :~  :*  %give  %fact  ~[/game/(scot %da u.game-id)/updates]
+                          %chess-update  !>([%undo-accepted u.game-id])
+                      ==
+                  ==
+              =,  u.game-state
+              =:
+                moves.game.u.game-state  ?:  =(+.ship-to-move our.bowl)
+                                           (snip (snip moves.game))
+                                         (snip moves.game)
+                position.u.game-state  ?:  =(+.ship-to-move our.bowl)
+                                         (fen-to-position (tail (rear (snip (snip moves.game)))))
+                                       (fen-to-position (tail (rear (snip moves.game))))
+                got-undo-request.u.game-state  |
+              ==
+              %=  this
+                games  (~(put by games) u.game-id u.game-state)
+              ==
           ==
       ==
   ==
@@ -1020,7 +1150,8 @@
   ?~  new-position
     [~ ~]
   =/  updated-game  `chess-game`game.game-state
-  =.  moves.updated-game  (snoc moves.updated-game move)
+  =/  fen  (position-to-fen u.new-position)
+  =.  moves.updated-game  (snoc moves.updated-game [move fen])
   =/  new-fen-repetition  (increment-repetition fen-repetition.game-state u.new-position)
   =/  in-checkmate  ~(in-checkmate with-position u.new-position)
   =/  in-stalemate  ?:  in-checkmate
--- a/src/urbit/lib/chess.hoon
+++ b/src/urbit/lib/chess.hoon
@@ -212,7 +212,7 @@
 +|  %rendering
 ::
 ::  render game result,
-::  which may or may not be null 
+::  which may or may not be null
 ++  result-string
   |=  result=(unit chess-result)
   ^-  @t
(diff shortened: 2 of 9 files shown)


src/urbit/sur/historic.hoon
```diff
@@ -0,0 +1,45 @@
+|%
+::
+::  chess-game stores metadata for a game
+::  represented by chess-position
++$  chess-game
```
ashelkovnykov: Should be `chess-game-0`, to denote when it was in the active state

src/urbit/app/chess.hoon
```diff
@@ -1,33 +1,51 @@
...
+      got-undo-request=?
   ==
-+$  state-0
++$  active-game-state-0
+  $:  game=chess-game.historic
+      position=chess-position
+      ready=?
+      sent-draw-offer=?
+      got-draw-offer=?
+  ==
```
ashelkovnykov: `active-game-state-0` should be in `historic.hoon` as well

src/urbit/app/chess.hoon
```diff
@@ -1,33 +1,51 @@
...
       games=(map @dau active-game-state)
       archive=(map @dau chess-game)
       challenges-sent=(map ship chess-challenge)
       challenges-received=(map ship chess-challenge)
       rng-state=(map ship chess-commitment)
   ==
++$  state-0
+  $:  %0
+      games=(map @dau active-game-state-0)
+      archive=(map @dau chess-game)
```
ashelkovnykov: This should be `chess-game-0.historic` (or possibly `chess-game-0`: see below comment)

src/urbit/app/chess.hoon
```diff
@@ -1,33 +1,51 @@
 ::  chess: fully decentralized, peer-to-peer chess app for urbit
 ::
 ::  import libraries and expose namespace
-/+  chess, dbug, default-agent
+/+  chess, dbug, default-agent, historic
 =,  chess
```
ashelkovnykov: Can add `=,  historic` below this line so that types don't need to be suffixed by `.historic`. This should work in all cases, as the historic types end with `-#` where `#` denotes in what game state they were active.

src/urbit/lib/chess.hoon
```diff
@@ -399,13 +399,121 @@
...
+      %none
+  ++  fen-to-en-passant
+    |=  fen=@t
+    ^-  (unit chess-square)
+    ?:  =(fen '-')
+      ~
+    =/  en-passant-square  (scan (trip fen) ;~(plug aln aln))
+    =/  file  `term`-.en-passant-square
+    =/  rank  `term`+.en-passant-square
+    `[file rank]
```
ashelkovnykov: This is a solid initial draft of parsing a FEN to a `chess-position`. However:
1. Pinning a named value to the subject and editing it is a natural response a lot of programmers have when coming to Hoon, but it's something that should be avoided as often as possible. The reason is that since Hoon is a functional language, it can't actually append or modify the subject/state: it can only clone existing state with modifications. This means that each state update is creating a copy of some subtree of the subject, which is inefficient, and thus should be avoided as much as possible.
2. Are you familiar with the `&n` and `|n` operators? `&n` produces the `nth` value of a tuple with at least `n + 1` entries. `|n` produces the value directly after `&n`. Example:
    ```
    > =y ['a' 'b' 'c']
    > &2.y
    'b'
    > |2.y
    'c'
    ```
3. It's a little odd that some things are parsed during splitting and others are parsed after splitting.
4. `parsed-fen` is somewhat of a misnomer, given that additional parsing happens on its data.

In addition to the above, there's some efficiency and correctness concerns that it would take a long time to patch going back and forth together. I recommend replacing the above implementation with this one:
```
++  fen-to-position
|=  fen=@t
^-  chess-position
%+  rush
  fen
|.
;~  (glue ace)
    fen-to-board         ::  board
    fen-to-player        ::  player-to-move
    fen-to-white-castle  ::  white-can-castle
    fen-to-black-castle  ::  black-can-castle
    fen-to-en-passant    ::  en-passant
    dem                  ::  ply-50-move-rule
    dem                  ::  move-number
==
++  fen-to-board
  %+  cook  board-helper
  %+  more  (just '/')
  ::  XX is this actually slower than alf and fail on invalid char?
  (stun [1 8] (mask "12345678bknpqrBKNPQR"))
++  fen-to-player
  ;~  pose
      (cold %white (just 'w'))
      (cold %black (just 'b'))
  ==
++  fen-to-white-castle
  ;~  pose
      (white-castle-helper hep " -" %none)
      (white-castle-helper (jest 'k') " k" %none)
      (white-castle-helper (jest 'q') " q" %none)
      (white-castle-helper (jest 'KQ') " " %both)
      (white-castle-helper (just 'K') " " %kingside)
      (white-castle-helper (just 'Q') " " %queenside)
  ==
++  fen-to-black-castle
  ;~  pose
      (cold %none hep)
      (cold %both (jest 'kq'))
      (cold %kingside (just 'k'))
      (cold %queenside (just 'q'))
  ==
++  fen-to-en-passant
  ;~  pose
      (cold ~ hep)
      %+  cook
          some
          ;~  plug
              (cook term (shim 'a' 'h'))
              (cook term (mask "36"))
          ==
  ==
++  board-helper
  |=  brd=(list tape)
  ?>  =((lent brd) 8)
  =/  rank  %8
  =/  file  %a
  =/  board  *(map chess-square chess-piece)
  |-
  ?~  brd
    board  
  ?<  (gth file %g)
  ?:  =(file %g)
    $(brd t.brd, rank (sub rank 1), file %a)
  ?+  (term  (cut 1 [(sub file 97) 3] i.brd))  !!
    %1  $(file (add file 1))
    %2  $(file (add file 2))
    %3  $(file (add file 3))
    %4  $(file (add file 4))
    %5  $(file (add file 5))
    %6  $(file (add file 6))
    %7  $(file (add file 7))
    %8  $(file (add file 8))
    ::
    %b  $(file +(file), board (~(put by board) [[file rank] [%black %bishop]]))
    %k  $(file +(file), board (~(put by board) [[file rank] [%black %king]]))
    %n  $(file +(file), board (~(put by board) [[file rank] [%black %knight]]))
    %p  $(file +(file), board (~(put by board) [[file rank] [%black %pawn]]))
    %q  $(file +(file), board (~(put by board) [[file rank] [%black %queen]]))
    %r  $(file +(file), board (~(put by board) [[file rank] [%black %rook]]))
    ::
    %B  $(file +(file), board (~(put by board) [[file rank] [%white %bishop]]))
    %K  $(file +(file), board (~(put by board) [[file rank] [%white %king]]))
    %N  $(file +(file), board (~(put by board) [[file rank] [%white %knight]]))
    %P  $(file +(file), board (~(put by board) [[file rank] [%white %pawn]]))
    %Q  $(file +(file), board (~(put by board) [[file rank] [%white %queen]]))
    %R  $(file +(file), board (~(put by board) [[file rank] [%white %rook]]))
  ==
++  white-castle-helper
  |=  [rul=rule pre=tape res=@tas]
  ^-  rule
  ;~(pfix rul (funk pre (easy res)))
```
Some notes about this implementation:
1. I didn't test it out, on anything: it's written entirely by intuition, just so that I don't take even longer on this PR and can get to the other ones. I'm hoping it's close enough to correct that I'm not laying a frustrating task onto you to patch it into its final state.
2. Despite catching more "bad FEN" cases, this is still an "unsafe" conversion function, since it doesn't check for a "well-formed" FEN. What we want is two `fen-to-position` functions: one safe, and one unsafe. The safe one can be used when we are converting FENs that we ourselves created back to `chess-position`s. The unsafe one is for importing FENs from externally. The unsafe version of the function should call the safe version, and then perform additional validation. Safe `fen-to-position` validation checks would include:
  a. exactly one king per side
  b. no more than 16 pieces per side
  c. no more than 10 rooks, bishops, or knights per side
  d. no more than 9 queens per side
  e. no more than 8 pawns per side
  f. no pawns on back ranks
  g. en passant square makes sense for board arrangement
  h. castle values make sense for board arrangement

Note that the safe version of `fen-to-position` is not necessary at this time. We can just add an empty arm with a comment about it, like `fen-to-position` had before this PR.

src/urbit/mar/chess/undo-accept.hoon
```diff
@@ -0,0 +1,11 @@
...
+++  grab
+  |%
+  ++  noun  ,~
+  --
+++  grow
+  |%
+  ++  noun  ~
+  --
+++  grad  %noun
+--
```
ashelkovnykov: Add a new line at the end of each mark file

src/urbit/app/chess.hoon
```diff
@@ -320,6 +339,93 @@
...
+                black.game.u.game-state
+            ==
+          ::  check whether it's our turn
+          ?>  ?=([%ship @p] ship-to-move)
+          ?:  (team:title +.ship-to-move src.bowl)
+            :_  this
+            =/  err
+              "opponent has already moved, can't request undo"
+            :~  [%give %poke-ack `~[leaf+err]]
+            ==
```
ashelkovnykov, line 444: This check should be relaxed: most often, the reason that someone requests an undo is because he makes a move that becomes apparent as a blunder after the opponent responds.

- If we're white, it's turn 25, it's our move, and we request an undo: we're trying to go back to our own move on turn 24.
- If we're white, it's turn 25, it's black's move, and we request an undo: we're tying to go back to our own move on turn 25.
- If we're black, it's turn 25, it's white's move, and we request an undo: we're trying to go back to our own move on turn 24.
- If we're black, it's turn 25, it's our move, and we request an undo: we're trying to go back to our own move on turn 24.

src/urbit/app/chess.hoon
```diff
@@ -320,6 +339,93 @@
...
+          ::  send undo request to opponent
+          :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
+                      %chess-undo-request  !>(~)
+                  ==
+              ==
```
ashelkovnykov, line 452: Need to send out an update on the `/game/../updates` wire so that the frontend can visualize the undo process

src/urbit/app/chess.hoon
```diff
@@ -320,6 +339,93 @@
...
+            :_  this
+            =/  err
+              "no undo request to decline for game {<game-id.action>}"
+              :~  [%give %poke-ack `~[leaf+err]]
+              ==
+          ::  decline undo request
+          :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
+                      %chess-undo-decline  !>(~)
+                  ==
+              ==
```
ashelkovnykov, line 479: Same here, re: `update` wire card

src/urbit/app/chess.hoon
```diff
@@ -320,6 +339,93 @@
...
+                  :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
+                      %chess-undo-accept  !>(~)
+                  ==
+              ==
+          =.  moves.game.u.game  (snip moves.game.u.game)
+          =.  position.u.game  (fen-to-position +.(rear moves.game.u.game))
+          =.  got-undo-request.u.game  |
+          %=  this
+            games  (~(put by games) game-id.action u.game)
+          ==
```
ashelkovnykov: Ugh - having to pin intermediate values to the subject just to update the state is inefficient, but there's no better way to do it here. I tried to write an alternative which updates the entire state in one go, and it's so hideously convoluted that it's probably better for the sake of readability to keep it like you wrote. Only recommendation I have is to merge the state updates using `=:` and to implement the undo functionality I described in one of the above comments:
```
=/  ship-to-move
  ?-  player-to-move.position.u.game-state
    %white  white.game.u.game-state
    %black  black.game.u.game-state
  ==
?>  ?=([%ship @p] ship-to-move)
...
=:
  moves.game.u.game  ?:(=(+.ship-to-move our) (snip moves.game.u.game) (snip (snip moves.game.u.game)))
  position.u.game (fen-to-position +.(rear moves.game.u.game))
  got-undo-request.u.game |
==
```

Unrelated:
1. These `ship-to-move` calls are really inefficient and unnecessary. This should just be part of the game state.
2. The way we handle moves is incredibly inefficient. We should be head-appending moves, instead of using `snoc`/`snip`. The frontend can reverse the list itself for in-progress games, otherwise it can build record moves in an array however it needs from moves along the `update` wire in a single session.

Those two points above are not actionable items for you, I'm just noting this down for posterity.

src/urbit/app/chess.hoon
```diff
@@ -766,6 +872,36 @@
...
+                games    (~(put by games) u.game-id u.game-state(sent-undo-request |))
+              ==
+            %chess-undo-accept
+              ?.  sent-undo-request.u.game-state
+                [~ this]
+              :-  :~  :*  %give  %fact  ~[/game/(scot %da u.game-id)/updates]
+                          %chess-update  !>([%undo-accepted u.game-id])
+                      ==
+                  ==
+              =,  game.u.game-state
```
ashelkovnykov: This isn't a bad idea, actually, if it were set to `=,  u.game-state`. This could be useful in the `%accept-undo` block above too: rename `game` to `game-state` and add this block after checking that `game-state` is non-null.

src/urbit/app/chess.hoon
```diff
@@ -766,6 +872,36 @@
...
+              =.  moves  (snip moves)
```
ashelkovnykov: Same advice as above, but invert the relationship based on whose turn it is. Combined with the preceding comment about using `=,`:
```
moves.game  ?:(=(+.ship-to-move our) (snip (snip moves.game)) (snip moves.game))
```

ashelkovnykov (changes requested):
What you have here is entirely workable - there are just some optimizations available to use that we should use.

src/urbit/lib/chess.hoon
rovmug-ticfyn: Based on [this post](https://developers.urbit.org/blog/text-overview), a term can't contain a capital letter, so we'll need a different way to differentiate the black and white pieces in `board-helper`. 

Maybe we first perform a `(sane @tas)` check. If yes, handle it with the `?+` you've written. If no, a series of `?:`s that check for the white pieces?

ashelkovnykov: Ooph - bummer, dude. Let me think on this!

src/urbit/lib/chess.hoon
```diff
@@ -399,13 +407,107 @@
...
+    %7  $(file (need (next-files [file 7])))
+    %8  $(file (need (next-files [file 8])))
+    ::
+    %b  $(file (need (next-file file)), board (~(put by board) [[file rank] [%black %bishop]]))
+    %k  $(file (need (next-file file)), board (~(put by board) [[file rank] [%black %king]]))
+    %n  $(file (need (next-file file)), board (~(put by board) [[file rank] [%black %knight]]))
+    %p  $(file (need (next-file file)), board (~(put by board) [[file rank] [%black %pawn]]))
+    %q  $(file (need (next-file file)), board (~(put by board) [[file rank] [%black %queen]]))
+    %r  $(file (need (next-file file)), board (~(put by board) [[file rank] [%black %rook]]))
+  ==
```
ashelkovnykov: Alright, here's a wild one for ya - replace these lines with:
```
?:  (lte next 56)
  $(file (chess-file (add file (sub next 48))))
?:  (gte next 97)
  ?+  (term next)  !!
    %b  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%black %bishop]]))
    %k  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%black %king]]))
    %n  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%black %knight]]))
    %p  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%black %pawn]]))
    %q  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%black %queen]]))
    %r  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%black %rook]]))
  ==
?+  (term (add next 32))  !!
  %b  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %bishop]]))
  %k  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %king]]))
  %n  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %knight]]))
  %p  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %pawn]]))
  %q  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %queen]]))
  %r  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %rook]]))
==
```
This is trying to use more [ASCII magic](https://www.asciitable.com) to convert a text character into a literal which Hoon can interpret as a type for `?+`. We also need to explicitly mold `file` with `chess-file` so that `next-file` isn't necessary (Look out for more comments about this).

src/urbit/lib/chess.hoon
```diff
@@ -399,13 +407,107 @@
...
+|.
+;~  (glue ace)
+    fen-to-board         ::  board
+    fen-to-player        ::  player-to-move
+    fen-to-white-castle  ::  white-can-castle
+    fen-to-black-castle  ::  black-can-castle
+    fen-to-en-passant    ::  en-passant
+    dem                  ::  ply-50-move-rule
+    dem                  ::  move-number
+==
```
ashelkovnykov: All of these lines should be indented once more

src/urbit/lib/chess.hoon
```diff
@@ -399,13 +407,107 @@
...
+  =/  rank=chess-rank  %8
+  =/  file=chess-file  %a
+  =/  board  *(map chess-square chess-piece)
+  |-
+  ?~  brd
+    board
+  ?<  (gth file %g)
+  ?:  =(file %g)
+    $(brd t.brd, rank (need (prev-rank rank)), file %a)
+  =*  next  (snag (sub file 97) i.brd)
```
ashelkovnykov: Okay, I made some mistakes in my first draft (including forgetting that `%g` comes *before* `%h`, not after it...). Try this:
```
=/  rank=chess-rank  %8
=/  file=$?(chess-file %i)  %a
...
  board
?-  file
    %i
  $(brd t.brd, rank (chess-rank (sub rank 1)), file %a)
::
    chess-file
  =*  next  (snag (sub file 97) i.brd)
  ...
==
```

ashelkovnykov (commented):
Okay, I think I've got some fixes for our parsing problem

ashelkovnykov:
@rovmug-ticfyn Whatever you did when adding your latest batch of changes has really confused GitHub. Could you crush all of your changes into one commit that's rebased directly onto the master branch? Hopefully that will fix it - otherwise, it's just impossible to review right now because GitHub is showing your changes intermixed with a ton of changes we've already merged.

src/urbit/sur/chess.hoon
```diff
@@ -210,8 +210,8 @@
     black=chess-player
   ::  %'0-1'
     result=(unit chess-result)
-  ::  a list of this round's moves
-    moves=(list chess-move)
+  ::  a list of this round's moves and corresponding fen
+    moves=(list [chess-move @t])
```
ashelkovnykov: We should probably declare a `chess-fen` type which just maps to `@t`, so that the type is self-documenting

src/urbit/sur/historic.hoon
```diff
@@ -0,0 +1,53 @@
...
+  ::  [%name 'Deep Blue']
+    white=chess-player
+  ::  [%name 'Garry Kasparov']
+    black=chess-player
+  ::  %'0-1'
+    result=(unit chess-result)
+  ::  a list of this round's moves
+    moves=(list chess-move)
+  ==  
+--
```
ashelkovnykov: Need newline at end of file

src/urbit/sur/historic.hoon
```diff
@@ -0,0 +1,53 @@
+/-  *chess
+|%
```
ashelkovnykov, line 2: Missing `state-0`

src/urbit/lib/chess.hoon
```diff
@@ -399,13 +407,103 @@
       ==
   --
 ::
-::  XX: implement or delete
+::  XX: need to implement
+::  turn fen into chess-position with checks for valid fen
+++  fen-to-position-safe
+|=  fen=@t
+~&  %not-implemented  !!
```
ashelkovnykov: These two lines should be indented two more

src/urbit/app/chess.hoon
```diff
@@ -1,33 +1,44 @@
...
       rng-state=(map ship chess-commitment)
   ==
++$  state-0
+  $:  %0
+      games=(map @dau active-game-state-0)
+      archive=(map @dau chess-game-0)
+      challenges-sent=(map ship chess-challenge)
+      challenges-received=(map ship chess-challenge)
+      rng-state=(map ship chess-commitment)
+  ==
```
ashelkovnykov: As commented above, should be in `historic.hoon`

src/urbit/app/chess.hoon
```diff
@@ -767,6 +866,38 @@
...
+              ?.  sent-undo-request.u.game-state
+                [~ this]
+              :-  :~  :*  %give  %fact  ~[/game/(scot %da u.game-id)/updates]
+                          %chess-update  !>([%undo-accepted u.game-id])
+                      ==
+                  ==
+              =,  u.game-state
+              =:
+                moves.game  ?:(=(+.ship-to-move our) (snip (snip moves.game)) (snip moves.game))
+                position  (fen-to-position (tail (rear moves.game)))
```
ashelkovnykov: Might be a bug here: just like we have to snip either two moves or one based on whose turn it is for `moves.game`, we need to get the FEN of either the most recent move or the second-most recent move

src/urbit/app/chess.hoon
```diff
@@ -321,6 +333,93 @@
...
+              ==
+          =/  ship-to-move
+            ?-  player-to-move.position
+              %white  white.game
+              %black  black.game
+            ==
+          ?>  ?=([%ship @p] ship-to-move)
+          =:
+            moves.game  ?:(=(+.ship-to-move our) (snip moves.game) (snip (snip moves.game)))
+            position  (fen-to-position (tail (rear moves.game)))
```
ashelkovnykov: Same as above

ashelkovnykov (changes requested):
Let me know if you want me to take a look at what's causing the `nest-fail` error

rovmug-ticfyn:
Getting a `nest-fail` error from the `fen-to-position` parsing telling me we have a nail that shouldn't be there (`[p=[p=@ud q=@ud] q=""]`). I'm still working on debugging, but no progress yet.

src/urbit/lib/chess.hoon
```diff
@@ -399,13 +407,103 @@
...
+++  fen-to-player
+  ;~  pose
+      (cold %white (just 'w'))
+      (cold %black (just 'b'))
+  ==
+++  fen-to-white-castle
+  ;~  pose
+      (white-castle-helper hep " -" %none)
+      (white-castle-helper (jest 'k') " k" %none)
+      (white-castle-helper (jest 'q') " q" %none)
```
ashelkovnykov: These two should be `just`, not `jest`

src/urbit/lib/chess.hoon
```diff
@@ -399,13 +407,103 @@
...
+    %b  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %bishop]]))
+    %k  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %king]]))
+    %n  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %knight]]))
+    %p  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %pawn]]))
+    %q  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %queen]]))
+    %r  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %rook]]))
+  ==
+==
+++  white-castle-helper
+  |=  [rul=rule pre=tape res=@tas]
```
ashelkovnykov: Should be `res=chess-castle`

src/urbit/lib/chess.hoon
```diff
@@ -399,13 +407,103 @@
...
+++  board-helper
+  |=  brd=(list tape)
+  ?>  =((lent brd) 8)
+  =/  rank=chess-rank  %8
+  =/  file=$?(chess-file %i)  %a
+  =/  board  *(map chess-square chess-piece)
+  |-
+  ?~  brd
+    board
+    ?-  file
```
ashelkovnykov: This line should be indented two spaces fewer, but with two more spaces before file:
```
  ?-    file
```

src/urbit/lib/chess.hoon
```diff
@@ -399,13 +407,103 @@
...
+    ==
+  ?+  (term (add next 32))  !!
+    %b  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %bishop]]))
+    %k  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %king]]))
+    %n  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %knight]]))
+    %p  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %pawn]]))
+    %q  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %queen]]))
+    %r  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %rook]]))
+  ==
+==
```
ashelkovnykov: These lines should be indented two spaces more

src/urbit/lib/chess.hoon
```diff
@@ -399,13 +407,103 @@
...
+              (cook term (shim 'a' 'h'))
+              (cook term (mask "36"))
+          ==
+  ==
+++  board-helper
+  |=  brd=(list tape)
+  ?>  =((lent brd) 8)
+  =/  rank=chess-rank  %8
+  =/  file=$?(chess-file %i)  %a
+  =/  board  *(map chess-square chess-piece)
```
ashelkovnykov: Should be `=/  board  *chess-board`

src/urbit/lib/chess.hoon
```diff
@@ -399,13 +407,103 @@
...
+    %n  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %knight]]))
+    %p  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %pawn]]))
+    %q  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %queen]]))
+    %r  $(file (chess-file +(file)), board (~(put by board) [[file rank] [%white %rook]]))
+  ==
+==
+++  white-castle-helper
+  |=  [rul=rule pre=tape res=@tas]
+  ^-  rule
+  ;~(pfix rul (funk pre (easy res)))
```
ashelkovnykov: This gate is the root cause of the `nest-fail` error - both uses of `rule` are too restrictive, so the compiler accidentally encodes very strict types for what a "rule" is at the point in time when we call it. Code should instead be:
```
|*  [rul=rule pre=tape res=chess-castle]
;~(pfix rul (funk pre (easy res)))
```

src/urbit/lib/chess.hoon
```diff
@@ -399,13 +407,103 @@
...
+      (cold %queenside (just 'q'))
+  ==
+++  fen-to-en-passant
+  ;~  pose
+      (cold ~ hep)
+      %+  cook
+          some
+          ;~  plug
+              (cook term (shim 'a' 'h'))
+              (cook term (mask "36"))
```
ashelkovnykov: Need explicit casts to `chess-file` and `chess-rank`:
```
(cook chess-file (cook term (shim 'a' 'h')))
(cook chess-rank (cook term (mask "36")))
```

src/urbit/lib/chess.hoon
```diff
@@ -36,6 +36,14 @@
...
   ?>  ?=(chess-file next)
   `next
+++  next-files
+  |=  [file=chess-file n=@ud]
+  ^-  (unit chess-file)
+  ?:  ?=(%h file)
+    ~
+  =/  next  (add file n)
+  ?>  ?=(chess-file next)
+  `next
```
ashelkovnykov: Not used

ashelkovnykov (changes requested):
Added comments which fix the `nest-fail` issue. Also, branch needs to be rebased onto most recent master. While looking at the PR, I made a branch that did both of these things, if you'd like to avoid repeating the same work. It's available [here](https://github.com/ashelkovnykov/urbit-chess/tree/undo).

Note that the code still does not compile after this - the `algebraicize` functions throw errors about missing the FEN (which we've added to the list of moves).

ashelkovnykov (commented):
Looks good to me. Stick a frontend on this and we're good!

ashelkovnykov (approved):
Approving PR to unblock other issues; frontend can come in a separate issue. Please squash commits before merge.

ashelkovnykov (approved):
LGTM

PR #52, opened 2022-10-12
Save and reset practice board state

### LINKED ISSUE #8: Save practice board state
The practice board configuration is lost when navigating away from it. Saving the state and adding a reset button would be helpful.

This could be done using the `%chess` app's state, but I think that's massively overkill. I'm thinking just using the browser's local storage, or a cookie. I'm not a frontend dev, so I don't know which one would be better.

Frontend work only.

rovmug-ticfyn:
Partially resolves #8 but currently has a bug that keeps the last move highlighted. Frontend button placement will likely need to change to fit UI changes in other issues.

### DIFF
--- a/src/frontend/tsx/Chessboard.tsx
+++ b/src/frontend/tsx/Chessboard.tsx
@@ -48,7 +48,7 @@ export function Chessboard () {
   const [chess, setChess] = useState<ChessInstance>(new Chess())
   const [promotionMove, setPromotionMove] = useState<PromotionMove | null>(null)
   const [renderWorkaround, forceRenderWorkaround] = useState<number>(Date.now())
-  const { urbit, displayGame, declinedDraw, offeredDraw, setDisplayGame } = useChessStore()
+  const { urbit, displayGame, declinedDraw, offeredDraw, setDisplayGame, practiceBoard, setPracticeBoard } = useChessStore()
 
   //
   // Non-state constants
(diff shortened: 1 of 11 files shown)


bonbud-macryg (changes requested):
The code here is solid, but I think you should slightly change the UI implementation.

One design principle for my proposed frontend work, which I might not have been clearly explicit about, is that we should be trying to make as little impact on the UI as we reasonably can. There'll be a UI redesign coming eventually, so we should just try to get our new functionality in without adding/changing more than we need to.

If you collapsed the "practice board" and "reset practice board" buttons into one, you'd reduce the impact and avoid an unnecessary button hanging around while the player is in a real game.

When the practice board is in its initial state, there should read "offer draw", "resign", and a grayed-out "reset practice board". When the practice board has had moves made on it, that grayed-out button should be active and ready to click. If the player clicks on a live game, that "reset practice board" button should then become the "practice board" button. You already have the `if (displayGame === null)` logic, so you can use that to determine the button's appearance/behaviour at any given time.

(Relatedly, the "offer draw" and "resign" buttons should be grayed-out and do nothing when the player is on the practice board.)

This is already a really good feature, the practice board is so much more useful now.

Once you've made those changes and you're confident it's ready to merge, you can save another round of review by squashing your commits in this PR into one, with a description summarising the changes, then rebasing that new commit to include the current state of `ashelkov:master`.

src/frontend/tsx/Chessboard.tsx
```diff
@@ -77,10 +77,21 @@ export function Chessboard () {
   const initBoard = () => {
     setApi(Chessground(boardRef.current, CHESSGROUND.baseConfig))
   }
+
```
bonbud-macryg: When I compiled with Docker it flagged up a "trailing spaces" error here; just clear the line so it's actually empty.

PR #53, opened 2022-10-18
Add webpack-only build workflow

bonbud-macryg:
There's good reason to keep npm etc. quarantined within Docker, but right now that workflow comes with a compile times around 100 seconds. Frontend developers need to make and test changes rapidly, and likely have these packages installed on their system already.

This commit adds a `webpack.config.nodocker.js` file and `build-webpack.sh` to easily build with it. Compiling with `build-webpack.sh` takes around 3 seconds.

A `package-lock.json` file is created by npm [for future developers (https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json#description) on any installs to manage dependencies, and I have no strong opinion on whether it should be removed from the GitHub repo. If developers have to install webpack and its plugins to run `build-webpack.sh`, a new `package-lock` will be created in their local `/src` folder.

This commit also modifies several frontend files to accommodate the new webpack workflow, but this doesn't affect the `build-docker` workflow.

### DIFF
--- a/src/frontend/package-lock.json
+++ b/src/frontend/package-lock.json
@@ -1,0 +1,10150 @@
+{
+  "name": "urbit-chess",
+  "version": "0.9.1",
+  "lockfileVersion": 2,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "urbit-chess",
+      "version": "0.9.1",
+      "license": "GPL3",
+      "dependencies": {
+        "@urbit/http-api": "^2.1.1",
+        "chess.js": "^0.13.3",
+        "chessground": "^8.2.1",
+        "react": "^18.2.0",
+        "react-beforeunload": "^2.5.3",
+        "react-dom": "^18.2.0",
+        "react-tabs": "^5.1.0",
+        "reactjs-popup": "^2.0.5",
+        "zustand": "^3.7.2"
+      },
+      "devDependencies": {
+        "@types/chess.js": "^0.13.1",
+        "@types/react": "^18.0.14",
+        "@types/react-beforeunload": "^2.1.1",
+        "@types/react-dom": "^18.0.5",
+        "@typescript-eslint/eslint-plugin": "^5.29.0",
+        "@typescript-eslint/parser": "^5.29.0",
+        "css-loader": "^6.7.1",
+        "eslint": "^8.18.0",
+        "eslint-config-ts-standard": "^1.0.5",
+        "eslint-plugin-import": "^2.26.0",
+        "eslint-plugin-node": "^11.1.0",
+        "eslint-plugin-promise": "^6.0.0",
+        "eslint-plugin-react": "^7.30.0",
+        "html-webpack-plugin": "^5.5.0",
+        "style-loader": "^3.3.1",
+        "ts-loader": "^9.3.0",
+        "typescript": "^4.7.4",
+        "url-loader": "^4.1.1",
+        "webpack": "^5.74.0",
+        "webpack-cli": "^4.10.0"
+      },
+      "engines": {
+        "node": ">=10 <19",
+        "npm": "^8.12.1"
+      }
+    },
+    "node_modules/@babel/code-frame": {
+      "version": "7.18.6",
+      "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.18.6.tgz",
+      "integrity": "sha512-TDCmlK5eOvH+eH7cdAFlNXeVJqWIQ7gW9tY1GJIpUtFb6CmjVyq2VM3u71bOyR8CRihcCgMUYoDNyLXao3+70Q==",
+      "dev": true,
+      "dependencies": {
+        "@babel/highlight": "^7.18.6"
+      },
+      "engines": {
+        "node": ">=6.9.0"
+      }
+    },
+    "node_modules/@babel/helper-validator-identifier": {
+      "version": "7.19.1",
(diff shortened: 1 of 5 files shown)


bin/build-docker.sh
```diff
@@ -134,4 +134,4 @@ sudo docker build --tag ${DOCKER_IMAGE}:${VERSION_FULL} .
 sudo docker run --rm -v ${FRONTEND_DIR}:/app/output/ ${DOCKER_IMAGE}:${VERSION_FULL}
 
 # Copy additional src files for frontend
-sudo chown -R ${USER}:${USER} ${FRONTEND_DIR}
+sudo chown -R ${USER}:${USER} ${FRONTEND_DIR}
```
ashelkovnykov: Please keep newline at end of script

src/frontend/package-lock.json
```diff
@@ -0,0 +1,10150 @@
+{
```
ashelkovnykov, line 1: 10,000 lines of dependencies - disgusting.

src/frontend/package.json
```diff
@@ -8,15 +8,24 @@
...
+    {
+      "name": "Raymond E. Pasco",
+      "email": "ray@ameretat.dev",
+      "url": "https://ameretat.dev"
+    },
+    {
+      "name": "Alex Shelkovnykov",
+      "email": "a@ashelkov.com",
+      "url": "https://ashelkov.com"
+    }
```
ashelkovnykov, line 20: Feel free to add yourself + @rovmug-ticfyn & @brbenji here

src/frontend/tsconfig.json
```diff
@@ -10,4 +10,4 @@
     "lib": ["ES2021", "DOM"],
     "esModuleInterop": true
   }
-}
+}
```
ashelkovnykov: Please keep newline at end of file

src/frontend/tsx/index.tsx
```diff
@@ -1,4 +1,9 @@
 import React from 'react'
+import '../css/buttons.css'
+import '../css/chessground.css'
+import '../css/general.css'
+import '../css/tabs.css'
+import '../css/textboxes.css'
```
ashelkovnykov, line 5: Any reason for adding these in? Should be taken care of by `index.html`

src/frontend/webpack.config.js
```diff
@@ -42,4 +42,4 @@ module.exports = {
      })
    ]
   }
-}
+}
```
ashelkovnykov: Please keep newline at end of file

bin/build-webpack.sh
```diff
@@ -0,0 +1,138 @@
...
+# Build desk.ship
+echo "~$SHIP" > $DESK_DIR/desk.ship
+
+# Build sys.kelvin
+echo "[%zuse $KELVIN]" > $DESK_DIR/sys.kelvin
+
+# Build frontend
+# sudo docker build --tag ${DOCKER_IMAGE}:${VERSION_FULL} .
+# sudo docker run --rm -v ${FRONTEND_DIR}:/app/output/ ${DOCKER_IMAGE}:${VERSION_FULL}
+(cd "$ROOT_DIR/src/frontend"; npm run build-no-docker)
```
ashelkovnykov: Looks like this is the only real change. So that we don't have two near identical scripts, each of which must have a correctly set version, let's replace this script with a `--no-docker` option in the regular build script:
```
...

usage() {
  if [[ $1 -ne 0 ]]; then
    exec 1>&2
  fi

  echo -e ""
  echo -e "Usage:\t$SCRIPT_NAME [-h] [-n] [-s SHIP_NAME] [-u URL]"
  echo -e ""
  echo -e "Build the app frontend and the desk files required to install it in Grid"
  echo -e ""
  echo -e "Options:"
  echo -e "  -h\tPrint script usage info"
  echo -e "  -n\tUse npm natively instead of through Docker"
  echo -e "  -s\tSet ship name to use (default: $DEFAULT_SHIP)"
  echo -e "  -u\tUse given URL to distribute glob over HTTP instead of over Ames"
  echo -e ""
  exit $1
}

...

DOCKER=1
DOCKER_IMAGE="urbit-chess"

...


OPTS=":hs:u:"
while getopts ${OPTS} opt; do
  case ${opt} in
    h)
      usage 0
      ;;
    n)
      DOCKER=0
    s)
      SHIP=$OPTARG
      ;;
    u)
      URL=$OPTARG
      ;;
    :)
      echo "$SCRIPT_NAME: Missing argument for '-${OPTARG}'" >&2
      usage 2
      ;;
    ?)
      echo "$SCRIPT_NAME: Invalid option '-${OPTARG}'" >&2
      usage 2
      ;;
  esac
done

...

# Build frontend
if [ $DOCKER -eq 1 ]; then
  sudo docker build --tag ${DOCKER_IMAGE}:${VERSION_FULL} .
  sudo docker run --rm -v ${FRONTEND_DIR}:/app/output/ ${DOCKER_IMAGE}:${VERSION_FULL}

  # Copy additional src files for frontend
  sudo chown -R ${USER}:${USER} ${FRONTEND_DIR}
else
  (cd "$ROOT_DIR/src/frontend"; npm run build-no-docker)
fi
```

ashelkovnykov (commented):
I, of course, refuse to put `npm` directly onto my machine, so I'll leave it to @rovmug-ticfyn or @brbenji to confirm whether the no-Docker build works.

src/frontend/tsx/index.tsx
bonbud-macryg: Huh, the webpack docs said they were necessary but both build workflows work without them. Removed.

ashelkovnykov (approved):
LGTM

PR #54, opened 2022-10-21
Game panel: integrating lichess/pgn-viewer (not ready)

brbenji:
### Two Commits for #37 

1. [add lichess-pgn-viewer with style](https://github.com/ashelkovnykov/urbit-chess/commit/01903de56b9b8460a856ec282463c3956d822d7b) is a proof of concept. This will only display a bare minimum pgn viewer. It doesn’t integrate at all with %chess. It required adding to webpack.config.js because of unspecified import paths.
2. [fail to integrate pgn-viewer into Chessboard](https://github.com/ashelkovnykov/urbit-chess/commit/e6d290d4602b4f25cd8887a846dd3518b28c38e7) is my best guess for utilizing lichess-pgn-viewer. I attempted to swap the Chessground function that controls our Chessboard with the LichessPgnViewer function. Currently, it doesn’t compile.

The [lichess-pgn-viewer](https://github.com/lichess-org/pgn-viewer) module has Chessground as a dependency, so I believe the use of it would mean the replacement of calling Chessground directly. The issue I ran into was with typing. The LichessPgnViewer default function is type void, which means it isn’t returning anything. That could be a fatal problem. I’ll continue to work on it.

### DIFF
--- a/src/frontend/css/lichess-pgn-viewer.css
+++ b/src/frontend/css/lichess-pgn-viewer.css
@@ -1,0 +1,717 @@
+@charset "UTF-8";
+.cg-wrap {
+  box-sizing: content-box;
+  position: relative;
+  display: block;
+  height: 0;
+  /*  padding-bottom: 100%;  breaks %chess chessboard layout */
+  width: 100%;
+}
+
+cg-container {
+  position: absolute;
+  width: 100%;
+  height: 100%;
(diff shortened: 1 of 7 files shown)


ashelkovnykov:
I think we all agreed that this is a rabbit hole not worth pursuing - closing the PR

PR #56, opened 2022-11-01
Add ability to save and reset practice board state to resolve #8

### LINKED ISSUE #8: Save practice board state

rovmug-ticfyn:


bonbud-macryg (approved):
I have a tiny adjustment on the button behaviour I'd like to see here but it shouldn't be a blocker, will discuss in the group. Otherwise LGTM.

PR #57, opened 2022-11-06
Add support for TypeScript -w compile

bonbud-macryg:
TypeScript's `tsc -w` command catches compile-time errors in your TypeScript files as soon as you save the file. This commit adds an "outDir" option for `tsconfig.json`, enabling a `tsc -w` compile.

This compile doesn't create a glob-ready `/build/frontend` folder. To enable that, we'd need another config file and maybe even another npm package. The `/build` folder should be cleared at the start of every compile before a new glob.

As is, the `tsc -w` compile is only for catching errors in your `.ts` and `.tsx` files as soon as you make them, rather than waiting for a "real" compile attempt to finish.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 build
 lila
 node_modules
+tsc-watch
--- a/src/frontend/tsconfig.json
+++ b/src/frontend/tsconfig.json
@@ -1,6 +1,6 @@
 {
   "compilerOptions": {
-    "outDir": "./dist/",
+    "outDir": "./tsc-watch",
     "noImplicitAny": true,
     "module": "ES2020",
     "target": "ES2021",


bonbud-macryg:
Should mention that `tsc -w` has to be run from `/src/frontend`.

bonbud-macryg:
Looking back at this now, all this PR really does is prevent a `src/dist` folder being created on every `tsc -w` build, at the cost of creating some confusion about what is and isn't glob-ready.

A better approach for now would be to set `outDir` to something like `./tsc-watch`, listed in `.gitignore`.

(npm requires this `tsconfig` file to build, but I've tested and `outDir` doesn't affect our `build.sh` compile.)

An ideal approach might be to one day support a compile workflow where `tsc -w` is the last step: once you're error-free, you can glob and test. This might require a third `config.json` file for npm to work from, one that just compiles the HTML and CSS and leaves the TypeScript to `tsc -w`. But for now, the first approach is enough.

ashelkovnykov:
I will leave frontend build changes entirely to @rovmug-ticfyn and @brbenji . So long as the changes don't affect the stability of the Docker workflow and the `build.sh` script, any of you three can merge such changes for each other after review approval.

brbenji:
`tsc -w` works as intended. It catches errors as you work, as well as when building with the “nodocker” webpack. and the docker build still works.

brbenji (approved):

PR #58, opened 2022-11-07
Add skeleton of new frontend

### LINKED ISSUE #37: Add panel to see previous moves, view previous board states

bonbud-macryg:
As we start to merge backend features, we need to be able to plug them into something resembling the Chess v1.0 frontend.

This commit partially implements the sketches we've discussed so far, just enough to keep us moving. None of the added UI is functional, added functionality will change the way it looks (button (un)selected statuses, etc.), and changes like moving buttons to the game panel are yet to be implemented.

The focus is on new buttons/menus corresponding to what we're working on over the next couple of weeks. Some minor visual tweaks have been made (e.g. capitalizing some UI text makes it more legible) but nothing here is remotely final.

### DIFF
--- a/src/frontend/css/general.css
+++ b/src/frontend/css/general.css
@@ -27,10 +27,12 @@ div.row {
 }
 
 .challenge-box {
+  display: flex;
+  align-items: center;
   justify-content: space-between;
 }
 
-.challenge-box .row {
+.challenge-box .row img {
   height: 60px;
   width: 60px;
 }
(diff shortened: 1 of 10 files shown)


src/frontend/css/general.css
```diff
@@ -96,41 +164,29 @@ div.board-container {
...
+  left: 5px;
+  top: -10px;
+  font-weight: normal;
 }
 
 ul.game-list {
   width: 100%;
   margin-top: 1rem;
-  border-top: 1px solid #666666;
+  padding-bottom: 1rem;
```
bonbud-macryg, line 208: This fixes an issue where the drop-shadow below the last game in a list would be cut off.

src/frontend/css/general.css
```diff
@@ -44,15 +44,82 @@ div.app-container {
...
 
+/* Game Panel */
+
+div.game-panel-container {
+  grid-area: game;
+  height: 100%;
+  width: 100%;
+  align-items: center;
+  justify-content: center;
+/*   border: red 1px solid; */
```
bonbud-macryg: I've left these `border` comments in for now as that styling is useful for figuring out visual issues.

src/frontend/css/general.css
```diff
@@ -44,15 +44,82 @@ div.app-container {
...
+/*   border: 1px solid yellow; */
+}
+
+div.game-panel #opp-timer {
+  justify-content: left;
+}
+
+div.game-panel #opp-player {
+  justify-content: left;
+}
```
bonbud-macryg, line 104: *As-is* these two CSS rules could be merged into `div.game-panel #opp-timer, #opp-player`, but I've kept everything separate for now as we might want to add styles later which fill these rules out and differentiate them.

src/frontend/css/general.css
```diff
@@ -176,3 +263,66 @@ div.challenge-side-container {
...
+
+div#challenges-header span {
+  cursor: pointer;
+}
+
+/* Settings menu */
+
+div.settings-container {
+  /* XX: overflow should be consistent for all menus */
```
bonbud-macryg, line 284: Disregard this.

src/frontend/tsx/Challenges.tsx
```diff
@@ -79,47 +79,62 @@ export function Challenges () {
...
+        {/* outgoing challenges list */}
+        <ul id="outgoing-challenges" className='game-list'>
+            {/* return outgoing list */}
+        </ul>
+        {/* mutual pals list */}
+        <ul id="pals" className="pals-list">
+            {/* return mutual pals list */}
+        </ul>
+      {/* new challenge pop-up */}
       {/* XX: should we host the LiChess images locally? */}
```
bonbud-macryg, line 132: Addressed in another issue, let's leave it for now.

src/frontend/css/general.css
ashelkovnykov: Big if true - well done

src/frontend/css/general.css
ashelkovnykov: Good idea. Maybe we should have our own thing in `%chess` like `XX` in Hoon code, as a signal that, "this is debug code and should be commented out unless you're actively looking for issues; please do not remove it". Maybe `DD`?

src/frontend/css/general.css
ashelkovnykov: Good idea, I agree

ashelkovnykov (commented):
- Various complaints about indentation from the linter
- The practice board is going to need either a separate game panel, or a separate flow within the game panel component
  - The timer and ship names are meaningless in the practice board
  - The plan is to move the game management buttons (e.g. "resign", "offer draw") to the game panel. This means that the "reset practice board" button is also going to need to move, almost certainly to the practice board's version of the game panel.
- The new incoming challenge card has buttons which are far too big. I understand that no user is going to have that many challenges at one time, but it just looks weird.
- Still refers to quick-challenge contacts as "mutuals" instead of "friends"
- Various visual bugs for oddly-shaped browser windows, which might occur if a user wants to have multiple windows open on his screen at one time. Examples: [quarter-sized](http://0x0.st/o6Kf.png), [half-sized](http://0x0.st/o6KV.png)

src/frontend/css/general.css
bonbud-macryg: Yeah that sounds good, I'll add it to the style guide.

ashelkovnykov (approved):
Yeah, this looks good. TBH, I didn't look at the code - I just loaded the frontend onto my fake-zod and played around with it.

It's good for building on top of, and we can take care of the minor visual bugs at the very end of the project.

desc

PR #59, opened 2022-11-10
Disable reset practice board button when in starting FEN

rovmug-ticfyn:


src/frontend/tsx/Games.tsx
```diff
@@ -41,13 +41,13 @@ export function Games () {
             disabled={!hasGame}
             onClick={() => setDisplayGame(null)}>
             practice board</button>
-         ) : (
+        ) : (
           <button
             className='option'
-            disabled={hasGame}
+            disabled={practiceBoard === 'rnbqkbnr/pppppppp/8/8/8/8/PPPPPPPP/RNBQKBNR w KQkq - 0 1'}
```
ashelkovnykov: 1. The default FEN already exists as a constant in `ts/constants/chess`
2. This is an incredibly inefficient way to determine if the button should be enabled or not. This is doing a string comparison every single time the state updates, which means it needs to compare every single character of the current FEN string to every single character of the default FEN string. The better way to do this would be to add a variable like `hasGame` named `practiceHasMoved` or something like that. When the practice board is reset, `practiceHasMoved` is set to `false`. If the user makes a move, `practiceHasMoved` is set to `true`. While `practiceHasMoved` is `false`, the "reset practice board" button is disabled.

bonbud-macryg:
@rovmug-ticfyn @ashelkovnykov The way this is implemented right now will conflict with #58, which moves the Practice Board buttons to the Game Panel.

rovmug-ticfyn:
@bonbud-macryg I’ll wait to refactor until the game panel is merged

bonbud-macryg:
@rovmug-ticfyn The skeleton of the game panel has been merged so you can refactor this whenever you're ready to.

We've talked in Design Notes about the buttons being icons rather than text like "reset practice board", but almost all of your refactored code should still apply there. Something to be aware of, but we shouldn't need to worry about those new buttons till the final UI tweaks.

ashelkovnykov (changes requested):
- `practiceHasMoved` and `setPracticeHaveMoved` shouldn't be in `ChessState`. They should probably be [in the React state, with a copy in `localStorage`](https://www.freecodecamp.org/news/how-to-use-localstorage-with-react-hooks-to-set-and-get-items/).
  - For that matter, the FEN state of the practice board should also be in `localStorage`. This would allow us to store the practice board state between looking at games and different browsing sessions.
- The "Reset Practice Board" button should be enabled/disabled based on `practiceHasMoved`, and hidden/visible based on `hasGame`

You should put these changes at the bottom of the priority stack, since they're not directly linked to any Milestone 1 issue.

ashelkovnykov (approved):
LGTM. I didn't check locally, so I won't hit the merge button. Anyone who's verified it's working locally (including yourself @rovmug-ticfyn) can go ahead and merge.

bonbud-macryg:
Happy to test it on my machine shortly, will merge if it's all good.

rovmug-ticfyn:
Not working as expected. Hold off on merging for now.

bonbud-macryg:
@rovmug-ticfyn Will wait until your say-so before I test it.

rovmug-ticfyn:
My latest commit causes the button to work as expected (disabled when the practice board is in starting position, enabled otherwise), but it does so by comparing the FEN string against the default each time the practice board is saved. I'm setting this aside for now and will come back to work on a more efficient solution.

src/frontend/tsx/GamePanel.tsx
```diff
@@ -6,6 +6,7 @@ import { Side, GameID, GameInfo, ActiveGameInfo } from '../ts/types/urbitChess'
 export function GamePanel () {
   const { urbit, displayGame, setDisplayGame, offeredDraw, practiceBoard, setPracticeBoard } = useChessStore()
   const hasGame: boolean = (displayGame !== null)
+  const practiceHasMoved = (localStorage.getItem('practiceHasMoved') !== null)
```
ashelkovnykov: Change this to:
```
const practiceHasMoved = (localStorage.getItem('practiceBoard') !== CHESS.defaultFEN)
```

src/frontend/tsx/GamePanel.tsx
```diff
@@ -96,7 +97,7 @@ export function GamePanel () {
         ) : (
           <button
             className='option'
-            disabled={hasGame}
+            disabled={!practiceHasMoved}
```
ashelkovnykov: Should be both the old condition and the new one:
```
disabled={hasGame || !practiceHasMoved}
```

src/frontend/tsx/Chessboard.tsx
```diff
@@ -197,6 +198,12 @@ export function Chessboard () {
     if (displayGame === null) {
       localStorage.setItem('practiceBoard', chess.fen())
       setPracticeBoard(chess.fen())
+      if (chess.fen() !== CHESS.defaultFEN) {
+        localStorage.setItem('practiceHasMoved', 'true')
+      }
+      else {
+        localStorage.removeItem('practiceHasMoved')
+      }
```
ashelkovnykov: You can drop all of the changes to this file

ashelkovnykov (changes requested):
Alright, so Chessground not being a React component makes this a burden. We basically have two options:
1. Do this the "right"/efficient way and have a separate boolean that records whether or not pieces on the practice board has moved. Another annoying thing is that changes to `localStorage` do not force a React re-render, so you need to do something like [what's outlined in this guide](https://www.tutorialspoint.com/localstorage-in-reactjs).
2. Just do the thing that I didn't want to do originally and that you're doing now: check if the FEN for the board is the initial FEN. This works every time, but it's inefficient/"wrong". However, dealing with it is so much easier that we should probably do it and write this off as one of those stupid things that is the cost of using JS (`~hastuc-dibtyx`'s Urbit-native UI can't come soon enough...).

ashelkovnykov (approved):
LGTM

PR #62, opened 2022-11-12
Fix a bug which kicked subscribers before sending result

ashelkovnykov:
This is the cause of situations where your opponent "never takes his turn" when a game-winning move is available.

### DIFF
--- a/src/urbit/app/chess.hoon
+++ b/src/urbit/app/chess.hoon
@@ -301,12 +301,13 @@
             :~  [%give %poke-ack `~[leaf+err]]
             ==
           =,  u.new.move-result
-          :-  %+  weld  cards.move-result
-              ^-  (list card)
+          :-  :_  cards.move-result
               ::  send move to opponent
-              :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
-                      %chess-move  !>(move.action)
-                  ==
+              :*  %give
+                  %fact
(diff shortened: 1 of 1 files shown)


src/urbit/app/chess.hoon
```diff
@@ -301,12 +301,13 @@
...
             ==
           =,  u.new.move-result
-          :-  %+  weld  cards.move-result
-              ^-  (list card)
+          :-  :_  cards.move-result
               ::  send move to opponent
-              :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
-                      %chess-move  !>(move.action)
-                  ==
+              :*  %give
```
rovmug-ticfyn, line 306: What's your reason for changing the style of the move card? If this is a preferred style, should someone go through and change existing cards for consistency?

ashelkovnykov: The old way was bothering me: I didn't like that the number of arguments was inconsistent per line, and it didn't match the Hoon style guide.

I can see what Ray was going for; the first three arguments and the last two arguments are split into "what Gall expects" and "what `%chess` expects", but it's just not a reliable system because of how long the subscriber wire could be.

At some point, all of the similar cards should have their style changed, but it's not an urgent issue that requires its own PR.

Unrelated: looks like the old "jogging"/"running" terminology for runes with variable input is gone?

rovmug-ticfyn (approved):
LGTM

Issue #67, opened 2022-11-26
Spectator mode


ashelkovnykov:
Since the frontend receives all of its updates by subscribing through the Urbit HTTP API, it would be fairly easy to enable live spectator mode by having other players subscribe to the same wire as the players use to interact with their game.

Features:
- invite to spectate
- generate `%chess` spectate link
- manual spectate mode

This should probably come after #63 and #64

PR #86, opened 2022-11-27
Add game panel functionality

### LINKED ISSUE #37: Add panel to see previous moves, view previous board states

bonbud-macryg:
@ashelkovnykov I'm just PR-ing this now to ask how you think I should update a game's move list upon receiving a new move. I've left two comments around line 60 in `chessStore.ts` where I think this should happen for an active game.

I'm not yet sure where I'm looking to update the practice board on a new move.

I've put the moves list in `%chess-game` on the backend and `GameInfo` on the frontend to account for archived games, which won't fall under `ActiveGameInfo`. But this creates some awkwardness in `receiveGame()` because I'm not getting new moves from the `%position` update itself. 

Sending a whole new list of moves in the `%position` update would require scrying `[%x %game @ta ~]` from `update.hoon`, which feels bad, and sending only the latest move would require:
* Some conditional logic on the frontend where the output of `algebraicize:with-position` (e.g. `'e4'`, `'d5'`) is appended to the array as `1. e4` or concatenated to get `1. e4 d5` depending on whether white or black moved.
* Amending the output of `algebraicize:with-position` from `'e4'` or `'d5'` to `'1. e4'` or `'1. e4 d5'`. That could comfortably be sent along with a `%position` update and appended to the `displayMoves` array without much complication. Alternatively I could write an arm `algebraicize-and-number:with-position` that does this. Either would let me append the move to the array in `receiveUpdate()` without much complication on the frontend. Considering you're thinking of splitting the frontend and backend, this is better than putting real game logic on the frontend. I've talked myself into this one in the process of writing it.

In either case, my second question is how to update `GameInfo` attributes from `receiveUpdate()`. The whole of `GameInfo` is updated in the line `info: currentGame.info`, but I couldn't do something like `info.moves = positionData.newMove`. I tried a whole `updateGameInfo` const that updated every attribute then changed that line to `info: updatedGameInfo`, but this crashed the whole frontend when I clicked on an active game. So maybe putting the moves list in `GameInfo` is just a bad idea, and I'll have to think of something else to account for both active and archived games.

### DIFF
--- a/src/frontend/ts/state/chessState.ts
+++ b/src/frontend/ts/state/chessState.ts
@@ -1,16 +1,20 @@
 import Urbit from '@urbit/http-api'
-import { Ship, GameID, GameInfo, ActiveGameInfo, Challenge, ChessUpdate, ChallengeUpdate } from '../types/urbitChess'
+import { Ship, GameID, SAN, FENPosition, Move, GameInfo, ActiveGameInfo, Challenge, ChessUpdate, ChallengeUpdate } from '../types/urbitChess'
 
 interface ChessState {
+  // properties
   urbit: Urbit | null
   displayGame: ActiveGameInfo | null
+  displayIndex: number | null
   practiceBoard: String | null
   activeGames: Map<GameID, ActiveGameInfo>
   incomingChallenges: Map<Ship, Challenge>
   outgoingChallenges: Map<Ship, Challenge>
   friends: Array<Ship>
+  // functions
   setUrbit: (urbit: Urbit) => void
   setDisplayGame: (displayGame: ActiveGameInfo | null) => void
+  setDisplayIndex: (displayIndex: number | null) => void
   setPracticeBoard: (practiceBoard: String | null) => void
   setFriends: (friends: Array<Ship>) => void
   receiveChallengeUpdate: (data: ChallengeUpdate) => void
--- a/src/frontend/ts/state/chessStore.ts
+++ b/src/frontend/ts/state/chessStore.ts
@@ -1,7 +1,7 @@
 import create from 'zustand'
 import Urbit from '@urbit/http-api'
 import { CHESS } from '../constants/chess'
-import { Update, Ship, GameID, GameInfo, ActiveGameInfo, Challenge, ChessUpdate, ChallengeUpdate, ChallengeSentUpdate, ChallengeReceivedUpdate, PositionUpdate, ResultUpdate, DrawOfferUpdate, DrawDeclinedUpdate, SpecialDrawPreferenceUpdate } from '../types/urbitChess'
+import { Update, Ship, GameID, SAN, FENPosition, Move, GameInfo, ActiveGameInfo, Challenge, ChessUpdate, ChallengeUpdate, ChallengeSentUpdate, ChallengeReceivedUpdate, PositionUpdate, ResultUpdate, DrawOfferUpdate, DrawDeclinedUpdate, SpecialDrawPreferenceUpdate } from '../types/urbitChess'
 import { findFriends } from '../helpers/urbitChess'
 import ChessState from './chessState'
 
@@ -13,10 +13,17 @@ const useChessStore = create<ChessState>((set, get) => ({
   incomingChallenges: new Map(),
   outgoingChallenges: new Map(),
   friends: [],
+  displayIndex: null,
   setUrbit: (urbit: Urbit) => set({ urbit }),
-  setDisplayGame: (displayGame: ActiveGameInfo | null) => set({ displayGame }),
+  setDisplayGame: (displayGame: ActiveGameInfo | null) => {
+    set({ displayGame, displayIndex: null })
+  },
   setPracticeBoard: (practiceBoard: String | null) => set({ practiceBoard }),
   setFriends: async (friends: Array<Ship>) => set({ friends }),
+  setDisplayIndex: (displayIndex: number | null) => {
+    set({ displayIndex })
+    console.log('setDisplayIndex displayIndex: ' + displayIndex)
+  },
   receiveChallengeUpdate: (data: ChallengeUpdate) => {
     switch (data.chessUpdate) {
       case Update.ChallengeSent: {
@@ -71,10 +78,8 @@ const useChessStore = create<ChessState>((set, get) => ({
   },
   receiveUpdate: (data: ChessUpdate) => {
     const updateDisplayGame = (updatedGame: ActiveGameInfo) => {
-      const displayGame = get().displayGame
-
-      if ((displayGame !== null) && (updatedGame.info.gameID === displayGame.info.gameID)) {
-        get().setDisplayGame(updatedGame)
+      if ((get().displayGame !== null) && (updatedGame.info.gameID === get().displayGame.info.gameID)) {
+        set({ displayGame: updatedGame, displayIndex: null })
       }
     }
 
@@ -82,24 +87,32 @@ const useChessStore = create<ChessState>((set, get) => ({
       case Update.Position: {
         const positionData = data as PositionUpdate
         const gameID = positionData.gameID
-        const specialDrawAvailable = positionData.specialDrawAvailable
-
+        const move = positionData.move
         const currentGame = get().activeGames.get(gameID)
-        const updatedGame: ActiveGameInfo = {
-          position: positionData.position,
-          gotDrawOffer: currentGame.gotDrawOffer,
-          sentDrawOffer: currentGame.sentDrawOffer,
-          drawClaimAvailable: specialDrawAvailable,
-          autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
-          info: currentGame.info
+
+        if (move.san !== null && move.fen !== null) {
+          currentGame.info.moves.push(move)
+
+          const updatedGame: ActiveGameInfo = {
+            position: move.fen,
+            gotDrawOffer: currentGame.gotDrawOffer,
+            sentDrawOffer: currentGame.sentDrawOffer,
+            drawClaimAvailable: positionData.specialDrawAvailable,
+            autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
+            info: currentGame.info
+          }
+
+          set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
+          updateDisplayGame(updatedGame)
+
+          console.log('RECEIVED POSITION UPDATE')
+          console.log('Update.Position displayIndex: ' + (get().displayIndex))
+          console.log('Update.Position fen: ' + move.fen)
         }
 
-        set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
-        updateDisplayGame(updatedGame)
-
-        console.log('RECEIVED POSITION UPDATE')
         break
       }
+
       case Update.Result: {
         const resultData = data as ResultUpdate
         const gameID = resultData.gameID
@@ -117,6 +130,7 @@ const useChessStore = create<ChessState>((set, get) => ({
         console.log('RECEIVED RESULT UPDATE')
         break
       }
+
       case Update.DrawOffer: {
         const offerData = data as DrawOfferUpdate
         const gameID = offerData.gameID
@@ -137,6 +151,7 @@ const useChessStore = create<ChessState>((set, get) => ({
         console.log('RECEIVED DRAW OFFER UPDATE')
         break
       }
+
       case Update.DrawDeclined: {
         const declineData = data as DrawDeclinedUpdate
         const gameID = declineData.gameID
--- a/src/frontend/ts/types/urbitChess.ts
+++ b/src/frontend/ts/types/urbitChess.ts
@@ -69,7 +69,14 @@ export type Ship = string
 
 export type GameID = string
 
+export type SAN = string
+
 export type FENPosition = string
+
+export type Move = {
+  san: SAN
+  fen: FENPosition
+}
 
 export type GameInfo = {
   gameID: GameID,
@@ -78,7 +85,8 @@ export type GameInfo = {
   round: string,
   white: Ship,
   black: Ship,
-  result: Result
+  result: Result,
+  moves: Array<Move>
 }
 
 export type ActiveGameInfo = {
@@ -133,6 +141,7 @@ export interface PositionUpdate extends ChessUpdate {
   gameID: GameID
   position: FENPosition
   specialDrawAvailable: boolean
+  move: Move | null
 }
 
 export interface ResultUpdate extends ChessUpdate {
--- a/src/frontend/tsx/App.tsx
+++ b/src/frontend/tsx/App.tsx
@@ -7,9 +7,10 @@ import { findFriends } from '../ts/helpers/urbitChess'
 import { Chessboard } from './Chessboard'
 import { Menu } from './Menu'
 import { GamePanel } from './GamePanel'
+import { PracticePanel } from './PracticePanel'
 
 export function App () {
-  const { urbit, setUrbit, receiveChallengeUpdate, receiveGame, setFriends } = useChessStore()
+  const { urbit, setUrbit, receiveChallengeUpdate, receiveGame, displayGame, setFriends } = useChessStore()
 
   //
   // Helper functions
@@ -56,7 +57,11 @@ export function App () {
   return (
     <Beforeunload onBeforeunload={teardown}>
       <div className='app-container'>
-        <GamePanel />
+        {
+          (displayGame == null)
+            ? <PracticePanel />
+            : <GamePanel />
+        }
         <Chessboard />
         <Menu />
       </div>
--- a/src/frontend/tsx/Chessboard.tsx
+++ b/src/frontend/tsx/Chessboard.tsx
@@ -7,14 +7,13 @@ import * as cg from 'chessground/types'
 import Popup from 'reactjs-popup'
 import { CHESS } from '../ts/constants/chess'
 import { CHESSGROUND } from '../ts/constants/chessground'
-import { URBIT_CHESS } from '../ts/constants/urbitChess'
 import { getChessDests, isChessPromotion } from '../ts/helpers/chess'
 import { getCgColor } from '../ts/helpers/chessground'
 import { pokeAction, move, castle, acceptDraw, declineDraw, claimSpecialDraw } from '../ts/helpers/urbitChess'
 import useChessStore from '../ts/state/chessStore'
 import usePreferenceStore from '../ts/state/preferenceStore'
 import { PromotionMove } from '../ts/types/chessground'
-import { Side, CastleSide, PromotionRole, Rank, File, GameID, ActiveGameInfo } from '../ts/types/urbitChess'
+import { Side, CastleSide, PromotionRole, Rank, File, GameID, GameInfo, ActiveGameInfo } from '../ts/types/urbitChess'
 
 //
 // Declare custom HTML elements used by Chessground
@@ -41,7 +40,7 @@ export function Chessboard () {
   const [chess, setChess] = useState<ChessInstance>(new Chess())
   const [promotionMove, setPromotionMove] = useState<PromotionMove | null>(null)
   const [renderWorkaround, forceRenderWorkaround] = useState<number>(Date.now())
-  const { urbit, displayGame, declinedDraw, offeredDraw, setDisplayGame, practiceBoard, setPracticeBoard } = useChessStore()
+  const { urbit, displayGame, declinedDraw, offeredDraw, setDisplayGame, practiceBoard, setPracticeBoard, displayIndex } = useChessStore()
   const { pieceTheme, boardTheme } = usePreferenceStore()
 
   //
@@ -63,6 +62,12 @@ export function Chessboard () {
       ? `${CHESS.pieceWhiteKnight} ${displayGame.info.white} vs. ${CHESS.pieceBlackKnight} ${displayGame.info.black}`
       : `${CHESS.pieceBlackKnight} ${displayGame.info.black} vs. ${CHESS.pieceWhiteKnight} ${displayGame.info.white}`
     : `~${window.ship}'s practice board`
+  const isViewOnly = (displayIndex == null)
+    ? false
+    : (displayGame.info.moves == null)
+      ? false
+      : (displayIndex < displayGame.info.moves.length - 1)
+  const toShowDests = !isViewOnly
 
   //
   // React hook helper functions
@@ -173,7 +178,9 @@ export function Chessboard () {
               attemptMove(orig, dest)
             }
 
-            forceRenderWorkaround(Date.now())
+            if (displayGame == null) {
+              forceRenderWorkaround(Date.now())
+            }
           }
         }
       }
@@ -183,13 +190,21 @@ export function Chessboard () {
 
   const updateBoard = () => {
     const stateConfig: CgConfig = {
-      fen: chess.fen(),
+      fen: (displayIndex == null || displayGame.info.moves == null)
+        ? chess.fen()
+        : displayGame.info.moves[displayIndex].fen,
+      viewOnly: isViewOnly,
       turnColor: sideToMove as cg.Color,
       check: chess.in_check(),
+      selected: null,
       movable: {
-        dests: getChessDests(chess) as cg.Dests
+        dests: getChessDests(chess) as cg.Dests,
+        showDests: toShowDests
       }
     }
+    // XX move these console logs to 'dev' branch once that exists
+    console.log('updateBoard fen: ' + stateConfig.fen)
+    console.log('updateBoard displayIndex: ' + displayIndex)
     api?.set(stateConfig)
   }
 
@@ -204,7 +219,9 @@ export function Chessboard () {
     if (practiceBoard === null) {
       localStorage.removeItem('practiceBoard')
       chess.load(CHESS.defaultFEN)
-      forceRenderWorkaround(Date.now())
+      if (displayGame == null) {
+        forceRenderWorkaround(Date.now())
+      }
       const config: CgConfig = {
         lastMove: null
       }
@@ -237,6 +254,12 @@ export function Chessboard () {
       updateBoard()
     },
     [displayGame])
+
+  useEffect(
+    () => {
+      updateBoard()
+    },
+    [displayIndex])
 
   useEffect(
     () => {
@@ -375,11 +398,6 @@ export function Chessboard () {
 
   return (
     <div className='game-container'>
-      <div className='title-container'>
-        <p className='title-text' style={{ fontSize: URBIT_CHESS.lengthToFontSize.get(boardTitle.length) }}>
-          {`${boardTitle}`}
-        </p>
-      </div>
       <div className={`board-container ${boardTheme} ${pieceTheme}`}>
         <div ref={boardRef} className='chessboard cg-wrap' />
         { (promotionMove !== null) ? renderPromotionInterface() : <div/> }
--- a/src/frontend/tsx/GamePanel.tsx
+++ b/src/frontend/tsx/GamePanel.tsx
@@ -1,13 +1,17 @@
-import React from 'react'
+import React, { useState } from 'react'
+import { Chess, ChessInstance } from 'chess.js'
 import useChessStore from '../ts/state/chessStore'
 import { pokeAction, resign, offerDraw, claimSpecialDraw } from '../ts/helpers/urbitChess'
-import { Side, GameID, GameInfo, ActiveGameInfo } from '../ts/types/urbitChess'
 import { CHESS } from '../ts/constants/chess'
+import { Side, GameID, SAN, GameInfo, ActiveGameInfo } from '../ts/types/urbitChess'
 
 export function GamePanel () {
-  const { urbit, displayGame, setDisplayGame, offeredDraw, practiceBoard, setPracticeBoard } = useChessStore()
+  const { urbit, displayGame, setDisplayGame, offeredDraw, practiceBoard, setPracticeBoard, displayIndex, setDisplayIndex } = useChessStore()
   const hasGame: boolean = (displayGame !== null)
   const practiceHasMoved = (localStorage.getItem('practiceBoard') !== CHESS.defaultFEN)
+  const opponent = !hasGame ? '~sampel-palnet' : (urbit.ship === displayGame.info.white.substring(1))
+    ? displayGame.info.black
+    : displayGame.info.white
 
   const resignOnClick = async () => {
     const gameID = displayGame.info.gameID
@@ -24,46 +28,67 @@ export function GamePanel () {
     await pokeAction(urbit, claimSpecialDraw(gameID))
   }
 
+  const moveOpacity = (index: number) => {
+    if (displayIndex == null || index <= displayIndex) {
+      return 1.0
+    } else {
+      return 0.3
+    }
+  }
+
+  const moveList = () => {
+    let displayMoves = (displayGame.info.moves !== null) ? displayGame.info.moves : []
+    let components = []
+    for (let wIndex: number = 0; wIndex < displayMoves.length; wIndex += 2) {
+      const move: number = (wIndex / 2) + 1
+      const bIndex: number = wIndex + 1
+      const wMove: SAN = displayMoves[wIndex].san
+
+      if (bIndex >= displayMoves.length) {
+        components.push(
+          <li key={ move } className='move-item' style={{ opacity: moveOpacity(wIndex) }}>
+            <span onClick={ () => setDisplayIndex(wIndex) }>
+              { wMove }
+            </span>
+          </li>
+        )
+      } else {
+        components.push(
+          <li key={ move } className='move-item' style={{ opacity: moveOpacity(wIndex) }}>
+            <span onClick={ () => setDisplayIndex(wIndex) }>
+              { wMove }
+            </span>
+            { '\xa0'.repeat(6 - wMove.length) }
+            {/* setting opacity to 1.0 offsets a cumulative reduction in opacity on each bIndex ply when displayIndex < this move's wIndex */}
+            <span onClick={ () => setDisplayIndex(bIndex) } style={{ opacity: (moveOpacity(wIndex) == 1.0) ? moveOpacity(bIndex) : 1.0 }}>
+              { displayMoves[wIndex + 1].san }
+            </span>
+          </li>
+        )
+      }
+    }
+
+    return components
+  }
+
   return (
-    <div className='game-panel-container col'>
+    <div className='game-panel-container col' style={{ display: ((displayGame !== null) ? 'flex' : ' none') }}>
       <div className="game-panel col">
-        <div id="opp-timer" className={'timer row' + (hasGame ? '' : ' invisible')}>
+        <div id="opp-timer" className={'timer row' + (hasGame ? '' : ' hidden')}>
           <p>00:00</p>
         </div>
-        <div id="opp-player" className={'player row' + (hasGame ? '' : ' invisible')}>
-          <p>~sampel-palnet</p>
+        <div id="opp-player" className={'player row' + (hasGame ? '' : ' hidden')}>
+          <p>{opponent}</p>
         </div>
-        <div className="moves col">
-          <div className="moves-divider"></div>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
-          <p> 00  ply ply</p>
+        <div className={'moves col' + (hasGame ? '' : ' hidden')}>
+          <ol>
+            { moveList() }
+          </ol>
         </div>
-        <div id="our-player" className={'player row' + (hasGame ? '' : ' invisible')}>
-          <p>~sampel-palnet</p>
+        <div id="our-player" className={'player row' + (hasGame ? '' : ' hidden')}>
+          <p>~{window.ship}</p>
         </div>
-        <div id="our-timer" className={'timer row' + (hasGame ? '' : ' invisible')}>
+        <div id="our-timer" className={'timer row' + (hasGame ? '' : ' hidden')}>
           <p>00:00</p>
         </div>
         {/* buttons */}
--- a/src/frontend/tsx/Games.tsx
+++ b/src/frontend/tsx/Games.tsx
@@ -4,7 +4,7 @@ import useChessStore from '../ts/state/chessStore'
 import usePreferenceStore from '../ts/state/preferenceStore'
 
 export function Games () {
-  const { urbit, displayGame, activeGames, setDisplayGame } = useChessStore()
+  const { urbit, displayGame, activeGames, setDisplayGame, setDisplayIndex } = useChessStore()
   const { pieceTheme } = usePreferenceStore()
   const hasGame: boolean = (displayGame !== null)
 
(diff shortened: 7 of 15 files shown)


ashelkovnykov (commented):
`receiveGame()` is called when an update is sent along the `/active-games` wire. This happens when a new game starts or when the frontend first subscribes to the backend (it receives one event for each active game).

I don't think you want to use `algebraicize-and-number` at all, since we want each ply of a move to be its own separate entry. It's not difficult to number the moves ourselves: every two entries in the `GameInfo` `Array` are one move, and get their own line with its own move number. The total number of moves is just `round-up(length-of-move-list / 2)`.

Note that there are two different `algebraicize` functions: one as its own arm in `lib/chess` and one as an arm in `with-position`. The former produces a list of moves in SAN using the latter.

You would never send the entire list of moves in a position update. You would only send a list of moves as a part of an entire `GameInfo` when the frontend first subscribes to the backend.

For active games that the frontend is already tracking, moves will arrive as `Position` type `ChessUpdate` objects. These will need to be extended to send the move which led to the position in SAN (unless it's an undo: then we just remove moves from the existing list).

To answer your bottom question, I think you would leave `info: currentGame.info` alone, and instead add the following line (once a move in SAN position is included in the update):
```
...
const currentGame = get().activeGames.get(gameID)
currentGame.info.moves.push(positionData.san)
const updatedGame: ActiveGameInfo = {
...
```

src/urbit/app/chess.hoon
```diff
@@ -627,7 +632,7 @@
       =/  fen  (position-to-fen position.game-state)
       =/  cards  ^-  (list card)
         :~  :*  %give  %fact  ~[/game/(scot %da u.game-id)/updates]
-                %chess-update  !>([%position u.game-id fen])
+                %chess-update  !>([%position u.game-id fen ''])
```
bonbud-macryg: The `''` here is where the `san` position update would go if there was one. Do I need to include this here? It would be nice if I didn't, as `useChessStore` currently needs some conditional logic to ignore a `''` position, so as not to start each game's `displayMoves` array with an empty index.

bonbud-macryg: It might be more correct (and frontend-agnostic) to change `''` to a `~` and check how `enjs:format` handles that, then update `useChessStore` accordingly.

src/frontend/ts/state/chessStore.ts
```diff
@@ -1,17 +1,37 @@
...
+      //     the "view completed games" feature
+      get().setDisplayMoves(displayGame.info.moves)
+    } else {
+      // XX: get moves list working for practice games
+      //
+      //     will require work in Chessboard.tsx, perhaps
+      //     work on the backend e.g. a new poke to just
+      //     get the PositionUpdate through /updates wire
+      set({ displayGame })
+      get().setDisplayMoves([])
```
ashelkovnykov: This is temporary, right? We're going to display the practice board moves as well?

src/urbit/app/chess.hoon
ashelkovnykov: Hmmm - if it's this annoying to deal with appending/removing moves in the frontend, maybe each position update should call `algebraicize-and-number` and return the entire list of moves, which can be stored in `ActiveGameInfo`? It's not the most efficient way to do this, but it might be significantly easier to code. I know I recommended against it earlier, but I didn't realize at the time how annoying it would be to manage the moves for each game in the frontend.

About your earlier comment, "Sending a whole new list of moves in the '%position' update would require scrying '[%x %game @ta ~]' from 'update.hoon'": no, you could call `(algebraicize-and-number game.game-state)` in `try-move` and `%chess-debug-inject`, and use the empty list (`~`) in `[%game @ta %updates ~]`.

ashelkovnykov (commented):
Think you may need to rebase onto the new master

src/urbit/app/chess.hoon
ashelkovnykov: Hmmm - but then after #47 goes in, that would need to be extended to sending the entire list of moves and their FENs after each new move, which is becoming really awful...

ashelkovnykov: So I guess the frontend will need a new type of object, which contains not just the `ActiveGameInfo`, but also an `Array` of move data, which is a pair of move SAN and FEN after that move. Getting a position update involves updating the `ActiveGameInfo` and adding another move to this `Array`. Accepting/receiving an undo involves removing a move from this `Array`.

bonbud-macryg: The problem with `algebraicize-and-number` is that it sends each cord like `'1. f3'` or `'1. f3 e6'`, so to make each ply clickable I'd need to parse and split each string as it hits the frontend to separate them into clickable `span`s. And we've no need for that arm's numbering of the list when we're currently doing it with a CSS rule.

bonbud-macryg: The specific concern I had was that `''` rather than `~` feels like a weird kludge. `~`s are returned throughout the app and that looks/feels fine. I'll amend it now and see how the frontend reacts.

ashelkovnykov: Yeah, ignore `algebraicize-and-number` -just send one SAN move at a time. To make this work for subscriptions from the frontend to in-progress games is that they'll need to send a stream of update cards for each move that has occurred so far, each containing the move SAN and FEN.

bonbud-macryg: That sounds good. The `''` to `~` change was fine so I'll revisit this issue once the undo functionality is merged, putting the SAN and FENs together in the updates.

src/frontend/ts/state/chessStore.ts
```diff
@@ -1,20 +1,40 @@
...
+    }
+  },
+  setDisplayMoves: (displayMoves: Array<SAN>) =>
+    set(state => ({ displayMoves })),
```
ashelkovnykov: If `setDisplayMoves` is only used from inside `ChessStore`, there isn't any reason to have it be its own function. May as well just replace both calls to it with `set(state => ({ displayGame.info.moves })` (or `[]`).

PS: Is `state` necessary there, or will `set({ displayGame.info.moves })` work on its own?

src/frontend/tsx/GamePanel.tsx
```diff
@@ -22,46 +25,46 @@ export function GamePanel () {
...
-          <p> 00  ply ply</p>
+        <div className={'moves col' + (hasGame ? '' : ' hidden')}>
+        <ol>
+        {
+          Array.from(displayMoves).map((ply, thisIndex, thisArray) => {
+            const nextIndex: number = thisIndex + 1
+            if (thisIndex % 2 === 0) {
+              return (
+              <li>
+                <span>{ply}</span> <span>{(nextIndex > thisArray.length ? '' : thisArray.at(nextIndex))}</span>
```
ashelkovnykov: This is still WIP, correct? Current code looks like it only works when both moves of a ply have been made.

src/urbit/app/chess.hoon
```diff
@@ -854,7 +857,7 @@
       =/  fen  (position-to-fen position.game-state)
       =/  cards  ^-  (list card)
         :~  :*  %give  %fact  ~[/game/(scot %da u.game-id)/updates]
-                %chess-update  !>([%position u.game-id fen special-draw-available.game-state])
+                %chess-update  !>([%position u.game-id fen special-draw-available.game-state ~])
```
ashelkovnykov: Is this card necessary? If we send the entire array of moves with the game info on start-up (each of which should come with its resulting FEN position), do we even need to send the latest position?

I guess we don't need it now, but it would be helpful for spectating mode, in the future (#67). Just thinking out loud here, not an issues to address.

src/frontend/tsx/GamePanel.tsx
bonbud-macryg: This is WIP until I request a proper review, yes. Since we're not making it work for the practice board anymore, the only thing left to do before review is get it browsing previous board states.

Right now the move list will correctly show a line like "01. e3" as soon as white has made their first move. If a second move hasn't been made, the conditional logic here will append nothing (`''`); if a second move has been made, it'll append the next move. This applies to the moves at index 0, 2, 4, 6 etc. in the array. There's no other logic around numbering/placing the moves in the UI, the numbering is just a CSS rule.

image

bonbud-macryg:
@ashelkovnykov I think this is ready for a review. It's pretty much *there*. Two known issues:
* The way I'm assigning `previousPosition` and `currentPosition` in `GamePanel.tsx` seems inefficient, but I'm not sure how to update just the one property of `displayGame`. We do something very similar for receiving position updates in `receiveUpdate`.
* If you click a piece to highlight possible moves, then click to review a previous position where it was your move, those moves will be highlighted on the board. Even though the relevant piece might not "line up" with the legal moves. We can't just switch the highlight off in `reviewMode` because we can't access `chess.js`.
  * Relatedly, the turn container will read "`your-side` to move..." when reviewing a previous position where it was your move. We *could* just make it invisible in `reviewMode`, but its behaviour flags up a possible cause:
  * Thinking about it now, the board is probably doing this because it's being told it's our move by either reading the FEN passed in, or taking in a `Side` property. This determines whether or not legal moves can/will be highlighted. If the latter, we could have `Side` be `null` whenever `reviewMode` is `true`. If the former, we shouldn't break the FEN by replacing `b` with `w` or vice versa, so what might we do?

src/urbit/mar/chess/update.hoon
```diff
@@ -47,6 +47,11 @@
             ['gameID' [%s (scot %da game-id.upd)]]
             ['position' [%s position.upd]]
             ['specialDrawAvailable' [%b special-draw-available.upd]]
+            :-  'move'
+            %-  pairs:enjs
+              :~  ['san' [%s san.upd]]
+                  ['fen' [%s position.upd]]
+              ==
```
ashelkovnykov: These lines should be indented two less

src/frontend/ts/constants/urbitChess.ts
```diff
@@ -1,54 +0,0 @@
...
-  [21, 'min(max(5.8vh, 48px), max(3.0vw, 54px))'],
-  [24, 'min(max(5.2vh, 48px), max(3.0vw, 46px))'],
-  [31, 'min(max(4.2vh, 48px), max(3.0vw, 36px))'],
-  [45, 'min(max(3.5vh, 48px), max(2.6vw, 25px))'],
-  [74, 'min(max(3.3vh, 48px), max(1.6vw, 18px))']
-])
-
-export const URBIT_CHESS = {
-  lengthToFontSize: LENGTH_TO_FONT_SIZE
-}
```
ashelkovnykov: Could you please leave this file in - I'm probably going to make use of it again. Making the chess interface reactive is a massive undertaking, so I'm probably going to make use of these values but adjust them based on the new minimum screen size.

src/frontend/ts/state/chessStore.ts
```diff
@@ -1,20 +1,40 @@
...
+      // XX: "GameInfo or ActiveGameInfo?" conditional logic
+      //
+      //     might be necessary depending on how we implement
+      //     the "view completed games" feature
+      get().setDisplayMoves([])
+      for (let move of displayGame.info.moves) {
+        get().displayMoves.push(move.san)
+      }
+    } else {
+      set({ displayGame })
```
ashelkovnykov: `set({ displayGame })` can be moved outside of the `if` statement, since both branches call it

src/frontend/ts/state/chessStore.ts
```diff
@@ -1,20 +1,40 @@
...
+    if (displayGame !== null) {
+      set({ displayGame })
+      // XX: "GameInfo or ActiveGameInfo?" conditional logic
+      //
+      //     might be necessary depending on how we implement
+      //     the "view completed games" feature
+      get().setDisplayMoves([])
+      for (let move of displayGame.info.moves) {
+        get().displayMoves.push(move.san)
+      }
```
ashelkovnykov: Active games shouldn't send their list of moves, we should do the other idea I brought up offline: we have another object in `ChessState` called `activeGameMoves` (or something like that) which is type `Map<GameID, Array<Move>>`. When the frontend first subscribes to the backend, the backend sends an `ActiveGameInfo` object without any moves, then it sends a `Position` update for each move in the game's state individually (so we'd generate one `%chess-update` card for each move in the active game in the `%active-games` branch of the `on-watch` arm).

`ChessUpdate`s with `Update.Position` type append the move SAN and FEN to the `Array` in the `Map` with the appropriate `GameID`. Then, to access SAN/FEN info for a game, we call `activeGameMoves.get(displayGame.gameID)`.

src/urbit/mar/chess/game.hoon
```diff
@@ -9,15 +10,23 @@
...
+        ['result' [%s ?~(result.game '' u.result.game)]]
+        :-  'moves'
+        :-  %a
+        %+  turn
+          moves.game
+        |=  move=[move=chess-move fen=chess-fen san=chess-san]
+        %-  pairs:enjs
+        :~  ['san' [%s san.move]]
+            ['fen' [%s fen.move]]
+        ==
```
ashelkovnykov: See above comment - this change shouldn't be necessary if we send a bunch of update cards for each game

src/urbit/app/chess.hoon
```diff
@@ -657,7 +657,10 @@
         ==
       =/  fen  (position-to-fen u.new-position)
       :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
-                  %chess-update  !>([%position game-id.action fen (check-50-move-rule u.new-position)])
+                  ::  XX: could replace ++rear of algebraicizing
+                  ::      whole move list with arm algebraicizing
+                  ::      just the one move
+                  %chess-update  !>([%position game-id.action fen (check-50-move-rule u.new-position) (rear (algebraicize game.action))])
```
ashelkovnykov: For this to work correctly with my above idea, we'd need a way to first dump the old moves and inject the new ones. As it stands right now, that would mean either:
1. Creating a new update card which will tell the frontend to dump state and sending that first. This is terrible in many ways, as it integrates frontend functionality with the backend.
2. Split the action into two pokes: first a the dev debugging would need to call `%chess-debug-zap`, then the dev would need to call `%chess-debug-inject`. This would still require adding a new update card to add the game to the list of active games.

Either way, having this stuff in the main code and needing to deal with it is becoming inconvenient: this is the second time making debug functionality work correctly is affecting a PR. I think that all of the debug code should be removed, and moved into a separate branch that lives permanently one commit ahead of master. This way, we can keep debug functionality out of the code that users actually use (which might protect them from exploits by other apps), but still allows devs to use it. Also, we can make a debug version of the frontend interface too, which would have convenience features, like debug zap/inject from the frontend. Creating a GH issue now - please just delete all of the debug pokes in this PR, and I'll add them in as a separate `debug` branch.

src/urbit/app/chess.hoon
ashelkovnykov: See new comment at this line in latest review

src/urbit/app/chess.hoon
```diff
@@ -835,7 +838,7 @@
         games  (~(put by games) u.game-id [new-game *chess-position *(map @t @ud) | & | | | | |])
       ==
     ::
-    ::  XX: document this
+    ::  subscribe to updates on active games
```
ashelkovnykov: Not what this does: this handles frontends subscribing to the chess agent to receive updates so that they can visualize them

src/urbit/app/chess.hoon
ashelkovnykov: With the implementation I outlined above, we would need to generate one card per move here, with both SAN and FEN

src/frontend/tsx/GamePanel.tsx
```diff
@@ -22,46 +26,77 @@ export function GamePanel () {
...
+              Array.from(displayMoves).map((ply, thisIndex, thisArray) => {
+                const nextIndex: number = thisIndex + 1
+                if (thisIndex % 2 === 0) {
+                  return (
+                    <li key={ thisIndex } className='move-item' style={{ opacity: (afterReviewIndex(thisIndex) ? 0.4 : 1.0) }}>
+                      <span onClick={() => reviewPosition(thisIndex)}>{ply}</span>{ '\xa0'.repeat(6 - ply.length) }<span onClick={() => reviewPosition(nextIndex)} style={{ opacity: (afterReviewIndex(thisIndex) ? 1.0 : afterReviewIndex(nextIndex) ? 0.4 : 1.0) }}>{(nextIndex > thisArray.length ? '' : thisArray.at(nextIndex))}</span>
+                    </li>
+                  )
+                }
+              })
```
ashelkovnykov: There's a more verbose but more elegant way to do this:
```
for (const i: number = 0; i < displayMoves.length; i += 2) {
  const move: number = (i / 2) + 1
  const wMove: SAN = displayMoves.get(i)
  const bIndex: number = i + 1

  if (bIndex > displayMoves.length) {

    return (
      <li key={ move } className='move-item'>
        <span onClick={() => setDisplayIndex(wIndex)} style={{ opacity: moveOpacity(wIndex) }>
          { wMove }
        </span>
      </li>
    )
  } else {
    return (
      <li key={ move } className='move-item'>
        <span onClick={() => setDisplayIndex(wIndex)} style={{ opacity: moveOpacity(wIndex) }>
          { wMove }
        </span>
        //
        { '\xa0'.repeat(6 - wMove.length) }
        //
        <span onClick={() => setDisplayIndex(bIndex)} style={{ opacity: moveOpacity(bIndex) }>
          { displayMoves.get(bIndex) }
        </span>
      </li>
    )
  }
}
```

Note that the above code fragment:
- Renames `reviewIndex`/`setReviewIndex` to `displayIndex`/`setDisplayIndex`
- Removes `reviewPosition`
- Rewrites `displayMoves` as a local constant which pulls the SAN data out of the `Array` of FEN-SAN pairs for the displayed game
- Replaces `afterReviewIndex` with the below function, which applies at the ply level now:
```
const moveOpacity = (index: number) => {
  (index > reviewIndex) : 0.4 ? 1.0
}
```

See my next comment regarding my thoughts on `reviewMode`/`reviewIndex`.

src/frontend/tsx/Chessboard.tsx
```diff
@@ -375,12 +374,7 @@ export function Chessboard () {
 
   return (
     <div className='game-container'>
-      <div className='title-container'>
-        <p className='title-text' style={{ fontSize: URBIT_CHESS.lengthToFontSize.get(boardTitle.length) }}>
-          {`${boardTitle}`}
-        </p>
-      </div>
-      <div className={`board-container ${boardTheme} ${pieceTheme}`}>
+      <div className={`board-container ${boardTheme} ${pieceTheme}`} style={{ pointerEvents: (reviewMode ? 'none' : 'auto') }}>
```
ashelkovnykov: `reviewMode` and messing around with the current position isn't what we should do. What we should be doing is:
1. Keep the current position, as far as the `chess` object (which implements all of the frontend chess logic) is concerned
2. Display the old position using Chessground
3. Make board `viewOnly` in the [Chessground configs](https://github.com/lichess-org/chessground/blob/master/src/config.ts) when browsing an old position

To accomplish 1, we just need to not touch the board position (which is why I deleted `reviewPosition` in the above comment).

To accomplish 2, the value for the `fen` Chessground config on line `185` should be: `fen: activeGameMoves.get(displayGame.gameId).get(displayIndex)`. (Note that this may require some fiddling, since the actual point of that line is to update the Chessground board to display the move that we just made inside of the `chess` object which actually records our chess state).

To accomplish 3, we would need to compute a new "non-state constant" (see line `50`) for whether the board is `viewOnly` (whether `displayIndex` is the final index of the `Array` of moves for this game or not). Then, we would set the `viewOnly` config in `updateBoard` just like we set the other configs (e.g. `fen`, `turnColor`, etc.).

Additional notes:
- Using the above implementation, we need to be mindful to increment `displayIndex` if it's at the latest move and we receive/make a new move.
- If we just started a game, the `Array` of `Move`s might be empty, so we need to handle the base case of a `displayIndex` of `0`.

ashelkovnykov (changes requested):

src/frontend/ts/state/chessStore.ts
ashelkovnykov: Display moves might not need to exist anymore, so I'm resolving this conversation

src/frontend/ts/constants/urbitChess.ts
```diff
@@ -51,4 +51,4 @@ const LENGTH_TO_FONT_SIZE: Map<number, string> = new Map([
 
 export const URBIT_CHESS = {
   lengthToFontSize: LENGTH_TO_FONT_SIZE
-}
+}
```
ashelkovnykov: Need to restore newline here

src/frontend/tsx/GamePanel.tsx
ashelkovnykov: @bonbud-macryg Looks like `for` loops can't return anything, so my code block above would need to be edited slightly:
```
let components = [];
for (const i: number = 0; i < displayMoves.length; i += 2) {
  ...
  if (bIndex > displayMoves.length) {
    components.push(...)
  } else {
    components.push(...)
  }
}
return components
```

ashelkovnykov (changes requested):

src/urbit/mar/chess/update.hoon
```diff
@@ -47,6 +47,11 @@
             ['gameID' [%s (scot %da game-id.upd)]]
             ['position' [%s position.upd]]
             ['specialDrawAvailable' [%b special-draw-available.upd]]
+            :-  'move'
+            %-  pairs:enjs
+            :~  ['san' [%s san.upd]]
+                ['fen' [%s position.upd]]
```
ashelkovnykov: You're sending the position twice right now: once here and once on line `48`

src/urbit/mar/chess/game.hoon
```diff
@@ -9,15 +10,15 @@
...
-        ['black' [%s (player-string:chess black)]]
-        ['result' [%s ?~(result '' u.result)]]
+    :~  ['gameID' [%s (scot %da game-id.game)]]
+        ['event' [%s event.game]]
+        ['site' [%s site.game]]
+        ['round' [%s (round-string:chess round.game)]]
+        ['white' [%s (player-string:chess white.game)]]
+        ['black' [%s (player-string:chess black.game)]]
+        ['result' [%s ?~(result.game '' u.result.game)]]
+        ['moves' [%a ~]]
```
ashelkovnykov, line 21: Is moves necessary here? I thought that the frontend would just cast the typeless object returned by `urbit/http-api` as a `GameInfo` and ignore missing fields or auto-populate them as `null`.

src/urbit/app/chess.hoon
```diff
@@ -853,10 +813,12 @@
...
       =/  cards  ^-  (list card)
-        :~  :*  %give  %fact  ~[/game/(scot %da u.game-id)/updates]
-                %chess-update  !>([%position u.game-id fen special-draw-available.game-state])
-            ==
-        ==
+         %+  turn
+           moves.game.game-state
+         |=  move=[move=chess-move fen=chess-fen san=chess-san]
+         :*  %give  %fact   ~[/game/(scot %da u.game-id)/updates]
+             %chess-update  !>([%position u.game-id fen.move | san.move])
```
ashelkovnykov: We're ignoring `special-draw-available.game-state` here (always returning the special draw state as false). Safe for now to just return whatever the current state is for every update, though it's not optimal for the correct function of interfaces.

src/frontend/tsx/Games.tsx
```diff
@@ -32,7 +32,7 @@ export function Games () {
                 key={key}
                 className={`game active ${colorClass} ${status}`}
                 title={gameID}
-                onClick={() => setDisplayGame(activeGame)}>
+                onClick={() => { setDisplayGame(activeGame); setDisplayIndex(activeGame.info.moves.length) }}>
```
ashelkovnykov: Array indices are 0-based, so `activeGame.info.moves.length` is actually one past the end of the Array

ashelkovnykov (changes requested):

src/urbit/app/chess.hoon
bonbud-macryg: I replaced `special-draw-available` with `|` because of the auto-claim setting. If player has auto-claim on, then receives a `%position` update where special draw is `&`, will the frontend end that game as it's being populated in the frontend? @rovmug-ticfyn

src/urbit/mar/chess/game.hoon
bonbud-macryg: I assumed that was the case, but I included `moves` thinking it was just stylistically better to explicitly send `~` than implicitly send nothing. Tested just now and turns out we do need it.

src/frontend/tsx/GamePanel.tsx
bonbud-macryg: This is kind-of working in the latest commit. I think the error is because it's enclosed in `Array.from` and it looks like the whole `for` loop should be replacing that. But as discussed, just putting it in curly brackets isn't accepted.

src/urbit/app/chess.hoon
rovmug-ticfyn: Ending the game with auto-claim happens entirely on the backend. The `special-draw-available` on the frontend only determines whether or not the Claim Special Draw button is enabled.

bonbud-macryg:
@ashelkovnykov One more UI bug. When I make a move in the practice board it animates correctly. When I make a move in a real game, the piece instantly appears on its destination square, then appears to pop back to the original square to begin the animation.

I think the bug is either in `updateBoard` in `Chessboard.tsx` or in the `Update.Position` branch of `chessStore.ts`. I suspect what's happening is that the board is receiving a new position twice. When I receive a move from the opponent, my board updates correctly, so maybe it's just a chessground issue that occurs when I click the destination square.

ashelkovnykov (commented):
Everything looks broadly correct in the code, though I'm experiencing bugs with the frontend:
1. Incoming challenges aren't being removed after being accepted (though outgoing challenges are)
2. Frontend crashes when attempting to open a game

Are you experiencing these too?

bonbud-macryg:
@ashelkovnykov Confirmed both bugs on my machine too. Frontend seems to crash when opening a *new* active game, existing games all function normally.

src/urbit/sur/chess.hoon
```diff
@@ -253,7 +257,7 @@
       [%challenge-received who=ship challenge=chess-challenge]
       [%challenge-resolved who=ship]
       [%challenge-replied who=ship]
-      [%position game-id=@dau position=@t special-draw-available=?]
+      [%position game-id=@dau position=@t special-draw-available=? san=@t]
```
ashelkovnykov: `san` should probably go before `special-draw-available`. I don't know if I can explain why - it's just one of "those things" about coding style: usually smaller/"more superfluous" members go towards the end. It might be rooted in the way structure members were laid out in C to reduce the overall structure size, based on the member alignments (how the member addresses were stored in memory), but I'm not sure.

src/urbit/mar/chess/game.hoon
ashelkovnykov: Once we're implementing browsing completed games, the `moves` object is where the FEN-SAN pairs for completed games will go, is that right?

src/frontend/tsx/GamePanel.tsx
```diff
@@ -24,46 +28,66 @@ export function GamePanel () {
...
+            { moveList() }
+          </ol>
         </div>
-        <div id="our-player" className={'player row' + (hasGame ? '' : ' invisible')}>
-          <p>~sampel-palnet</p>
+        <div id="our-player" className={'player row' + (hasGame ? '' : ' hidden')}>
+          <p>~{window.ship}</p>
         </div>
-        <div id="our-timer" className={'timer row' + (hasGame ? '' : ' invisible')}>
+        <div id="our-timer" className={'timer row' + (hasGame ? '' : ' hidden')}>
```
ashelkovnykov, line 91: We should probably make a separate "Practice Panel" that contains the "Reset Practice Board" button and can do it's own thing. For now, it would just have that button in it.

When a game against a human opponent is active, the entire "Practice Panel" would be hidden, and the "Game Panel" would be visible. Otherwise, it would be vice versa. The old, old version of the chess frontend used to do this. I think the code is as simple as:
```
if (activeGame) {
  <GamePanel />
} else {
  <PracticePanel />
}
```

src/frontend/tsx/Chessboard.tsx
```diff
@@ -183,13 +194,20 @@ export function Chessboard () {
...
       check: chess.in_check(),
+      selected: null,
       movable: {
-        dests: getChessDests(chess) as cg.Dests
+        dests: getChessDests(chess) as cg.Dests,
+        showDests: toShowDests
       }
     }
+    console.log('updateBoard fen: ' + stateConfig.fen)
+    console.log('updateBoard displayIndex: ' + displayIndex)
```
ashelkovnykov, line 207: Keep these for now, and we'll move them to the 'dev' branch once that exists.

src/frontend/ts/types/urbitChess.ts
```diff
@@ -133,6 +141,7 @@ export interface PositionUpdate extends ChessUpdate {
   gameID: GameID
   position: FENPosition
   specialDrawAvailable: boolean
+  move: Move | null
```
ashelkovnykov, line 144: Sorry, it's been a while - when is it possible for this to be `null`?

src/frontend/ts/state/chessStore.ts
```diff
@@ -197,6 +239,10 @@ const useChessStore = create<ChessState>((set, get) => ({
     updatedGame.sentDrawOffer = true
 
     set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
+  },
+  setDisplayIndex: (displayIndex: number | null) => {
+    set({ displayIndex })
+    console.log('setDisplayIndex displayIndex: ' + displayIndex)
```
ashelkovnykov: Please move up to be with the other "set" functions

src/frontend/ts/state/chessStore.ts
ashelkovnykov: Why not use `activeGameMoves` here?

src/frontend/ts/state/chessStore.ts
ashelkovnykov: Reopening pending resolution to above comment

src/frontend/ts/state/chessState.ts
```diff
@@ -1,23 +1,29 @@
...
   setDisplayGame: (displayGame: ActiveGameInfo | null) => void
+  setDisplayMoves: (displayMoves: Array<SAN> | null) => void
   setPracticeBoard: (practiceBoard: String | null) => void
   setFriends: (friends: Array<Ship>) => void
   receiveChallengeUpdate: (data: ChallengeUpdate) => void
   receiveGame: (data: GameInfo) => void
   receiveUpdate: (data: ChessUpdate) => void
   declinedDraw: (gameID: GameID) => void
   offeredDraw: (gameID: GameID) => void
+  setDisplayIndex: (displayIndex: number | null) => void
```
ashelkovnykov: Please move up to be with the other "set" functions

src/frontend/ts/state/chessState.ts
```diff
@@ -1,23 +1,29 @@
...
   displayGame: ActiveGameInfo | null
+  displayMoves: Array<SAN> | null
   practiceBoard: String | null
   activeGames: Map<GameID, ActiveGameInfo>
+  activeGameMoves: Map<GameID, Array<Move>>
   incomingChallenges: Map<Ship, Challenge>
   outgoingChallenges: Map<Ship, Challenge>
   friends: Array<Ship>
+  displayIndex: number | null
+
```
ashelkovnykov: Is this space meant to delimit members from functions?

src/frontend/ts/state/chessState.ts
```diff
@@ -1,23 +1,29 @@
...
   urbit: Urbit | null
   displayGame: ActiveGameInfo | null
+  displayMoves: Array<SAN> | null
   practiceBoard: String | null
   activeGames: Map<GameID, ActiveGameInfo>
+  activeGameMoves: Map<GameID, Array<Move>>
   incomingChallenges: Map<Ship, Challenge>
   outgoingChallenges: Map<Ship, Challenge>
   friends: Array<Ship>
+  displayIndex: number | null
```
ashelkovnykov: `displayIndex` should probably be grouped with `displayGame` and `displayMoves`

ashelkovnykov (changes requested):
Just tried it out - everything worked great!

Minor changes requested

src/frontend/ts/state/chessState.ts
bonbud-macryg: Yeah, but looking at it now it's a bit of a Hoonish thing to do here; I'll break it up with comments like a normal dev.

src/frontend/ts/types/urbitChess.ts
bonbud-macryg: I'm sure there was a good reason for this that wasn't "the app doesn't load unless this is allowed to be null", but I tried removing `| null` and turns out the app doesn't load unless this is allowed to be null.

Every time I've had this "the whole app crashes/doesn't load" error it's because it's trying to display something that isn't there. In this case the app is doing something as soon as I open it that causes this error.

Seems like when you open the frontend it subscribes to the `/updates` wire, then the `on-watch` endpoint sends over a `%position` update for every move that's been made in the game. If no moves have been made in that game, then the only update being sent has `null` where a move should be.

This is the relevant code from `on-watch`:
```
=/  cards  ^-  (list card)
   %+  turn
     moves.game.game-state
   |=  move=[move=chess-move fen=chess-fen san=chess-san]
   :*  %give  %fact   ~[/game/(scot %da u.game-id)/updates]
       %chess-update  !>([%position u.game-id fen.move special-draw-available.game-state san.move])
   ==
```

I've tried a couple of variations of "if `moves` is null then don't send anything", which cause issues on the frontend or backend. Could be that something has to give: we either *have to deal with* the value (with `| null`) or we *have to deal with* the whole game/cards/wire behaviour.

src/urbit/mar/chess/game.hoon
bonbud-macryg: Yeah that sounds right.

src/frontend/ts/state/chessStore.ts
bonbud-macryg: I think the thinking was:
1. We're setting `displayGame` here, so it feels more appropriate use `displayGame`'s list of moves.
2. It might be that not every `displayGame` being set here is an active game; they might be completed games. As gestured at in the comment here this function could change depending how #6 is implemented.

src/frontend/ts/types/urbitChess.ts
ashelkovnykov: That makes sense - letting it be `null` is definitely the easier solution. I think @rovmug-ticfyn has something similar in his PR for when someone tries to undo the first move.

src/urbit/mar/chess/update.hoon
```diff
@@ -45,12 +45,12 @@
         %-  pairs:enjs
         :~  ['chessUpdate' [%s 'position']]
             ['gameID' [%s (scot %da game-id.upd)]]
-            ['specialDrawAvailable' [%b special-draw-available.upd]]
             :-  'move'
             %-  pairs:enjs
             :~  ['san' [%s san.upd]]
                 ['fen' [%s position.upd]]
             ==
+            ['specialDrawAvailable' [%b special-draw-available.upd]]
```
ashelkovnykov, line 49: I don't think order matters for JSON, so you can reorder these however looks nicer if you'd like

src/frontend/tsx/App.tsx
```diff
@@ -57,6 +58,7 @@ export function App () {
     <Beforeunload onBeforeunload={teardown}>
       <div className='app-container'>
         <GamePanel />
+        <PracticePanel />
```
ashelkovnykov: It would be better to explicitly use `displayGame` here, either as input to the `GamePanel` and `PracticePanel`, or as the conditional to an `if` or `switch` statement. This makes it clear at the `App` level that only one of these should ever be rendered, not both:
```
<div className='app-container'>
  {
    (displayGame == null)
      ? <PracticePanel />
      : <GamePanel />
  }
  <Chessboard />
  <Menu />
</div>
```

src/frontend/ts/state/chessStore.ts
ashelkovnykov: I think we're too deep in the weeds here - I'll link what I'm getting at here shortly.

ashelkovnykov (changes requested):
One more comment incoming, in the form of a commit

src/frontend/ts/state/chessStore.ts
ashelkovnykov: We can get rid of `displayMoves` and `activeGameMoves` completely, since we're not using them. We'll have to account for this later, when dealing with archived games, but it won't be too bad since JS is garbage that allows hacky solutions to the hacky problems it creates. We can probably declare `displayGame` to be a union of `ActiveGameInfo` and `GameInfo` and just cast it depending on whether we're rendering an active or archived game.

Take a look at [this](https://github.com/ashelkovnykov/urbit-chess/commit/85e630c2e98152dcefd7615e28245463889a9c61) commit - it's what I was trying to get at with my past few comments about this specific block of code. It took 5 minutes to write and 2 hours to test (ty JS).

bonbud-macryg: Just Cc'ing @brbenji about the above comment.

src/frontend/tsx/GamePanel.tsx
```diff
@@ -24,46 +28,66 @@ export function GamePanel () {
...
+          </li>
+        )
+      } else {
+        components.push(
+          <li key={ move } className='move-item' style={{ opacity: moveOpacity(wIndex) }}>
+            <span onClick={ () => setDisplayIndex(wIndex) }>
+              { wMove }
+            </span>
+            { '\xa0'.repeat(6 - wMove.length) }
+            <span onClick={ () => setDisplayIndex(bIndex) } style={{ opacity: (moveOpacity(wIndex) == 0.3) ? 1.0 : moveOpacity(bIndex) }}>
```
ashelkovnykov: We should add a comment here that opacity ratings are cumulative, because it took me some time to understand why we were flipping the values here.

Duplicating specific hard values like this is an anti-pattern. Consider the case where we decide we want to change the opacity to `0.4` instead: now we have to change a specific value in two places in the code. The two common solutions are: 1) give the value a name, such as `moveOpacity`, and use the name instead, so that the value is only assigned once; 2) Change the way the value is assigned so that a single location in a function assigns it.

For the above reason, I would recommend swapping the ternary operator condition and arms:
```
opacity: (moveOpacity(wIndex) == 1.0) ? moveOpactiy(bIndex) : 1.0
```
This is more acceptable, since 100% is a natural default opacity (though if we ever wanted to change that, we'd now need named constants or a new `moveOpacity` function).

src/frontend/ts/state/chessState.ts
```diff
@@ -20,4 +24,4 @@ interface ChessState {
   offeredDraw: (gameID: GameID) => void
 }
 
-export default ChessState
+export default ChessState
```
ashelkovnykov: Add newline to end of file

src/frontend/ts/state/chessStore.ts
```diff
@@ -200,4 +215,4 @@ const useChessStore = create<ChessState>((set, get) => ({
   }
 }))
 
-export default useChessStore
+export default useChessStore
```
ashelkovnykov: Here too

src/frontend/tsx/App.tsx
```diff
@@ -56,12 +57,16 @@ export function App () {
...
+        }
         <Chessboard />
         <Menu />
       </div>
     </Beforeunload>
   )
 }
 
-export default App
+export default App
```
ashelkovnykov: Here too

src/frontend/tsx/Chessboard.tsx
```diff
@@ -390,4 +408,4 @@ export function Chessboard () {
       { (displayGame !== null) ? renderDrawPopup(displayGame) : <div/> }
     </div>
   )
-}
+}
```
ashelkovnykov: Here too

src/frontend/tsx/GamePanel.tsx
```diff
@@ -105,4 +129,4 @@ export function GamePanel () {
       </div>
     </div>
   )
-}
+}
```
ashelkovnykov: Here too

src/frontend/tsx/Games.tsx
```diff
@@ -53,4 +53,4 @@ export function Games () {
       </ul>
     </div>
   )
-}
+}
```
ashelkovnykov: Here too

ashelkovnykov (changes requested):
Final round of feedback

src/frontend/tsx/GamePanel.tsx
bonbud-macryg: Ah yeah, this is much more elegant.

ashelkovnykov (approved):
LGTM

PR #96, opened 2022-12-11
Refactor the Challenges menu

bonbud-macryg:
Connects the Incoming, Outgoing, and Friends lists into one coherent Challenges menu.

To fix a UI bug with the Friends list, it's been refactored to be a "first-class" piece of the app state along with the Incoming and Outgoing challenge lists. The `/friends` wire is now scried when the player first opens the app.


ashelkovnykov (approved):
LGTM

PR #102, opened 2023-01-02
Fix incomingChallenges bug

bonbud-macryg:
Fixes bug where incomingChallenges wouldn't update when accepting a challenge.

### DIFF
--- a/src/urbit/app/chess.hoon
+++ b/src/urbit/app/chess.hoon
@@ -615,6 +615,9 @@
                     :*  %give  %fact  ~[/active-games]
                         %chess-game  !>(new-game)
                     ==
+                    :*  %give  %fact   ~[/challenges]
+                        %chess-update  !>([%challenge-replied src.bowl])
+                    ==
                 ==
             %=  this
               challenges-received  (~(del by challenges-received) src.bowl)


bonbud-macryg:
I don't think this is the most elegant solution. It doesn't solve the fundamental problem that, if you put `console.log` in `chessStore`'s `challengeReplied` update, `challengeReplied` only fires when a challenge is rejected, but not accepted. But backend code in `/app` and `/mar` looks correct and I couldn't spot a mistake in the frontend last night. Might take another look at this again later; next step might be to just put a sigpam just above the `challenge-replied` cards to check they're both firing on the backend, then place more sigpams to check they're coming through the mark correctly.

ashelkovnykov:
@bonbud-macryg Yeah, this isn't solving the root problem.

I just spent an hour poking around, trying to figure out what's wrong, since everything looks okay. I tried comparing master to the state of the code right after merge, but I couldn't find anything that looked significant. At one point, I thought I was going mad because *sometimes* it would work correctly. Than I found the reproducible error case: **this only happens when the sides are randomized**.

This is because the `%accept-game` poke has a separate path for randomized sides: it doesn't immediately start a new game, instead the two players have to do a second handshake to determine who plays which side. I double-checked the original commit, and apparently we *never* accounted for this, so this bug has been present the entire time; we just missed it, somehow.

Therefore, the real solution is to send a `%chess-replied` poke for separately for this case. We have two places where it might go:
1. After line `119`, along with the initial `%chess-rng` poke
2. After line `617`, which is more symmetric (remove challenge from interface after user has clicked 'accept' and sides have been correctly assigned)

I think 2 is more correct, since that's also when the challenge is removed from the backend `set`.

bonbud-macryg:
Good find. (So *that's* what `%chess-rng` does...) I think 2 sounds more intuitive, I'll get on this sometime this week.

bonbud-macryg:
@ashelkovnykov Added the card and tested that it works with White, Random, and Black challenges.

ashelkovnykov (approved):
LGTM

Issue #65, opened 2022-11-26
Search games archive


ashelkovnykov:
Search set of games in archive to find specific a game. Search criteria ideas:
- opponent
- side
- date
- result

Blocked by #6

brbenji:
We could add 
- event

This would allow people to do a simple kind of tagging.

To be exhaustive of what’s available in a `chess-game`:
- site
_I can see this being used to name whatever frontend version of %chess is used. Or probably more useful would be if chess tournaments could change this. Ex: `site='Battle Chess 2023!`_ 
- round
_This one is a stretch, but perhaps it’s useful to view all the games you lost or won within 20 moves._

brbenji:
@ashelkovnykov I have an efficiency question.

Since the structure of the chess archive is `archive=(map @dau chess-game)` and `chess-game` contains all the relevant search info, will it be inefficient for a search to unpack every single `chess-game` in the `archive` map, gather up what matches the search, and then send the result?

I wouldn’t have thought about it except that there’s a cheat for date data, because we can just search the `@dau` directly in the `archive`, and send the matching `chess-game`s. 

**to start**
For the sake of just getting started, I’ll assume that %chess will send the whole `archive` to the frontend on a `peek` and that’s where we’ll handle searching through it.

ashelkovnykov:
@brbenji

I originally titled this issue "Search games archive", but I think that this may be a misnomer, depending on how we want to approach the heart of the request.

What do we actually want to do? We have archived games in `%chess` and we want to be able to browse them from the frontend. We'd also like to be able to filter the full list of games by one or more criteria, to make finding the one which interests use easier. Depending on how many games we have played and how many filters are applied, there may be too many to display at one time, so we want to browse them a few at a time.

The criteria which we would like to use for filtering (at least to begin with) are:
- Date
- Opponent
- Side played
- Result

"Site", "event", and "round" are not long for this world and should be ignored (this type of metadata is better handled outside of the app, unless tournaments have built-in support). Your tagging idea is a good idea, but shouldn't be included in the first draft. "Event" is already being used as a description, and we'll probably want to keep that (though rename it). One possibility is to use the description as a makeshift tag (i.e. filter by name) before a generic tagging system is implemented.

It's important to note that the starting date & time of the game is a natural ordering, and that we would never want to re-order the data: just filter it.

There are two ways we could accomplish this:
1. "Filter" approach
[...]

PR #109, opened 2023-01-27
Browse completed games

brbenji:
Screen Shot 2023-01-27 at 4 13 54 PM
This resolves #6 and works the way you'd expect.
_Now resolving #2 with the latest commit._

1. Clicking the Archive button will scry `archive` in %chess state.
_currently this just scries for all archives using path `archive/all`_
2. Clicking on a specific archive game will populate the GamePanel with moves and remove the draw and resign buttons.
_this adds a design bug for the GamePanel which is justified in the center of its container_

The biggest additions are the new `ArchiveGameInfo` type and the `%chess-archive` mark.

And the biggest modification is the `displayGame`, which is now a type union of `ActiveGameInfo | ArchiveGameInfo`.
The `ArchiveGameInfo` has superfluous parameters for draw offers, etc. But it seemed best for being able to use the existing `displayGame` infrastructure. 

Also, in making the `%chess-archive` mark I added a few helpful things to `sur/` and `lib/`.

PR #112, opened 2023-02-01
Archive search

brbenji:
**For discussion of archive searching #65 and converting of archive into a mop.**
Only look over: urbit file changes.
I'll comment where we need to discuss.

### DIFF
--- a/src/urbit/app/chess.hoon
+++ b/src/urbit/app/chess.hoon
@@ -445,9 +445,6 @@
           :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
                       %chess-undo-request  !>(~)
                   ==
-                  :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
-                      %chess-update  !>([%undo-request game-id.action])
-                  ==
               ==
           ::  record that undo has been requested
           %=  this
@@ -472,9 +469,6 @@
           :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
                       %chess-undo-decline  !>(~)
                   ==
-                  :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
-                      %chess-update  !>([%undo-declined game-id.action])
-                  ==
               ==
           %=  this
             ::  record that undo request is gone
@@ -494,17 +488,8 @@
           ?.  got-undo-request
             :_  this
             =/  err
-              "no undo request to decline for game {<game-id.action>}"
+              "no undo request to accept for game {<game-id.action>}"
               :~  [%give %poke-ack `~[leaf+err]]
-              ==
-          ::  tell opponent we accept the undo request
-          :-  :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
-                      %chess-undo-accept  !>(~)
-                  ==
-                  ::  update observers that we accept the undo request
-                  :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
-                      %chess-update  !>([%undo-accepted game-id.action])
-                  ==
               ==
           =/  ship-to-move
             ?-  player-to-move.position.u.game-state
@@ -521,8 +506,18 @@
                                    (fen-to-position (head (tail (rear (snip (snip moves.game))))))
             got-undo-request.u.game-state  |
           ==
-          %=  this
-            games  (~(put by games) game-id.action u.game-state)
+          :_  %=  this
+                games  (~(put by games) game-id.action u.game-state)
+              ==
+          ::  tell opponent we accept the undo request
+          :~  :*  %give  %fact  ~[/game/(scot %da game-id.action)/moves]
+                  %chess-undo-accept  !>(~)
+              ==
+              ::  update observers that we accept the undo request
+              :*  %give  %fact  ~[/game/(scot %da game-id.action)/updates]
+                  %chess-update
+                  !>([%undo-accepted game-id.action (position-to-fen position.u.game-state) ?:(=(+.ship-to-move our.bowl) ~.1 ~.2)])
+              ==
           ==
       ==
     ::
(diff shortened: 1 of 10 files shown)


src/urbit/app/chess.hoon
```diff
@@ -25,7 +25,7 @@
 +$  state-1
   $:  %1
       games=(map @dau active-game-state)
-      archive=(map @dau chess-game)
+      archive=((mop @dau chess-game) lth)
```
brbenji: do we think we'll need to make anything else in state a `mop`?

src/urbit/app/chess.hoon
```diff
@@ -834,49 +834,82 @@
...
+        [%all ~]
+      ``[%chess-archive !>(archive)]
+    ::
+    ::  =sur -build-file /=chess=/sur/chess/hoon
+    ::  .^(chess-archive:sur %gx /=chess=/archive/between/<game id>/<game id>/noun)
+    ::  .^(json %gx /=chess=/archive/between/<game id>/<game id>/json)
+        [%between @ta @ta ~]
+      =/  start=(unit @dau)  `(unit @dau)`(slaw %da i.t.t.path)
+      =/  end=(unit @dau)  `(unit @dau)`(slaw %da i.t.t.path)
+      ``[%chess-archive !>((lot:arch-orm archive start end))]
```
brbenji: relevant for #65 

while this compiles just fine and returns a `mop`, I don't think the `lot:on` is actually excluding any games outside the `@ta` starts and ends I pick.  also, it doesn't seem to be ordered either*. 

_I'll have to come back to it. maybe collapsing the `mop` to a `list` in the %chess-archive mark is causing problems?_

src/urbit/lib/chess.hoon
```diff
@@ -1394,6 +1395,33 @@
       (algebraicize-and-number game)
     ==
 ::
+::  helps chess-archive mark
+++  archive-json
+  |=  archive=[@dau chess-game]
+  =/  game  +.archive
```
brbenji: This is the extra logic added for the `archive` as `mop` that I'm not sure would transfer to the `%chess-game` mark.
Because collapsing a `mop` into a list will often make a `(list [key value])`, we have to only grab the tail when converting to json.

In PR #109, I bring it up in [this comment](https://github.com/thecommons-urbit/chess/pull/109#discussion_r1093451608)

src/urbit/mar/chess/archive.hoon
```diff
@@ -0,0 +1,19 @@
...
+  |%
+  ++  noun  chess-archive
+  --
+++  grow
+  |%
+  ++  noun  archive
+  ++  json
+    %-  frond:enjs
+    :-  'localArchive'
+    :-  %a  (turn (tap:arch-orm archive) archive-json:chess)
```
brbenji: collapsing things into a list might be causing the ordered map problems, _like not ordering things_.

brbenji (commented):
Here are a few things worth discussing for issue #65 and making choices in PR #109.

src/urbit/app/chess.hoon
bonbud-macryg: See [here](https://github.com/thecommons-urbit/chess/pull/109#discussion_r1095046515).

brbenji: Go [there](https://github.com/thecommons-urbit/chess/pull/109#discussion_r1095046515).

ashelkovnykov: Should be resolved by [this comment](https://github.com/thecommons-urbit/chess/issues/65#issuecomment-1424056224) on #65; please reopen conversation if that's not the case.

src/urbit/app/chess.hoon
ashelkovnykov: See #65; if we use the frontend filter approach, you can handle this from the frontend by walking over the ordered list of metadata and only inserting entries that meet all of the filter criteria into a display set that gets rendered under the archives tab.

src/urbit/lib/chess.hoon
ashelkovnykov: Based on my thoughts in #65, this will need to be slightly different: we should send the metadata info but not the moves when sending the archive.

I'd also prefer it to remain in the `mark` file. I don't think there's anything incorrect  about how you chose to do it, it just seems to be a convention that mark files go to great lengths to be self-contained. It's also a very standard JSON conversion, so I don't think it warrants being in the `lib`.

src/urbit/mar/chess/archive.hoon
ashelkovnykov: If the `tap` function is resulting in incorrect order from an ordered map, that's a pretty big bug...

Can you verify that you're getting things out of order, or produce a toy test example that works on any ship?

ashelkovnykov:
I think the way forward should be pretty clear after reading #65, but let me know if you have questions. A few thoughts:
- Look into React packages for doing automatic pagination. I'm sure there are JS libraries for automatically displaying lists of items that can't fit within some space as pages. This would make it really simple to display many archived games: instead of needing to manually track which results we're displaying and where we left off in the list, we just run a filter pass over the full list of archived game metadata, then throw it into the pagination component.
- Take a look at the `%challenges` and `%active-games` wires in `on-watch` for the best way to send a list of state. We should do the same for sending the state of archived games: one update per game, in order. We reassemble the list (in order) on the front end as an array. This will also make it easy for when games are completed: remove game from active games, add game to archived games, send update to frontend. Boom - it's now immediately searcable.
- I didn't look at much of the code in this PR, but it should be fine because I think you have a good idea of what needs to change, given the architectural direction change outlined in #65.
- I think you can safely close this PR, moving the code into #109, and then editing it there. We can continue the rest of the discussion re: game archives there.

src/urbit/mar/chess/archive.hoon
brbenji: I found my user error.

In [app/chess](https://github.com/thecommons-urbit/chess/pull/112/files#diff-b795e051f49f72ee01409845fe30b201053ae7e7e4e6a6712aade843f305c196) `~(put by archive)` wasn't replaced with `(put:arch-orm archive )` yet.

strange, but good, that the map arms work for mop. 

I also expected the archive to reorganize itself when I tweaked the mop or even the archive mark, but apparently the way it's `put` in "locks" things in place.

PR #118, opened 2023-02-25
Fix last move highlighting to resolve #36

### LINKED ISSUE #36: Most recent move not highlighted on game board
I believe this bug is caused by combining Chessground with React. Since `%chess` currently hard-resets the board to the new FEN and redraws the board after every move, it loses the highlighting of the previous move. Hopefully there's a graceful way to force highlighting the previous move.

rovmug-ticfyn:
My thinking on this is that any kind of state involving the last move squares should come from the backend so that we don't have the potential of the frontend highlighting getting out of sync with the backend. 

Currently getting an error converting from chess-update to json, so it seems there's something wrong with how I've written the array conversion but I can't tell from the docs.

### DIFF
--- a/src/frontend/ts/state/chessStore.ts
+++ b/src/frontend/ts/state/chessStore.ts
@@ -65,8 +65,7 @@ const useChessStore = create<ChessState>((set, get) => ({
       autoClaimSpecialDraws: false,
       gotUndoRequest: false,
       sentUndoRequest: false,
-      info: data,
-      lastMove: null
+      info: data
     }
 
     set(state => ({ activeGames: state.activeGames.set(data.gameID, activeGame) }))
@@ -104,8 +103,7 @@ const useChessStore = create<ChessState>((set, get) => ({
             autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
             gotUndoRequest: currentGame.gotUndoRequest,
             sentUndoRequest: currentGame.sentUndoRequest,
-            info: currentGame.info,
-            lastMove: positionData.lastMove
+            info: currentGame.info
           }
 
           set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
@@ -150,8 +148,7 @@ const useChessStore = create<ChessState>((set, get) => ({
           autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
           gotUndoRequest: currentGame.gotUndoRequest,
           sentUndoRequest: currentGame.sentUndoRequest,
-          info: currentGame.info,
-          lastMove: currentGame.lastMove
+          info: currentGame.info
         }
 
         set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
@@ -174,8 +171,7 @@ const useChessStore = create<ChessState>((set, get) => ({
           autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
           gotUndoRequest: currentGame.gotUndoRequest,
           sentUndoRequest: currentGame.sentUndoRequest,
-          info: currentGame.info,
-          lastMove: currentGame.lastMove
+          info: currentGame.info
         }
 
         set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
@@ -198,8 +194,7 @@ const useChessStore = create<ChessState>((set, get) => ({
           autoClaimSpecialDraws: setting,
           gotUndoRequest: currentGame.gotUndoRequest,
           sentUndoRequest: currentGame.sentUndoRequest,
-          info: currentGame.info,
-          lastMove: currentGame.lastMove
+          info: currentGame.info
         }
 
         set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
@@ -221,8 +216,7 @@ const useChessStore = create<ChessState>((set, get) => ({
           autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
           gotUndoRequest: true,
           sentUndoRequest: currentGame.sentUndoRequest,
-          info: currentGame.info,
-          lastMove: currentGame.lastMove
+          info: currentGame.info
         }
 
         set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
@@ -244,8 +238,7 @@ const useChessStore = create<ChessState>((set, get) => ({
           autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
           gotUndoRequest: currentGame.gotUndoRequest,
           sentUndoRequest: false,
-          info: currentGame.info,
-          lastMove: currentGame.lastMove
+          info: currentGame.info
         }
 
         set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
@@ -268,8 +261,7 @@ const useChessStore = create<ChessState>((set, get) => ({
           autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
           gotUndoRequest: false,
           sentUndoRequest: false,
-          info: currentGame.info,
-          lastMove: currentGame.lastMove
+          info: currentGame.info
         }
 
         set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
--- a/src/frontend/ts/state/chessStore.ts
+++ b/src/frontend/ts/state/chessStore.ts
@@ -65,7 +65,8 @@ const useChessStore = create<ChessState>((set, get) => ({
       autoClaimSpecialDraws: false,
       gotUndoRequest: false,
       sentUndoRequest: false,
-      info: data
+      info: data,
+      lastMove: null
     }
 
     set(state => ({ activeGames: state.activeGames.set(data.gameID, activeGame) }))
@@ -103,7 +104,8 @@ const useChessStore = create<ChessState>((set, get) => ({
             autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
             gotUndoRequest: currentGame.gotUndoRequest,
             sentUndoRequest: currentGame.sentUndoRequest,
-            info: currentGame.info
+            info: currentGame.info,
+            lastMove: positionData.lastMove
           }
 
           set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
@@ -148,7 +150,8 @@ const useChessStore = create<ChessState>((set, get) => ({
           autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
           gotUndoRequest: currentGame.gotUndoRequest,
           sentUndoRequest: currentGame.sentUndoRequest,
-          info: currentGame.info
+          info: currentGame.info,
+          lastMove: currentGame.lastMove
         }
 
         set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
@@ -171,7 +174,8 @@ const useChessStore = create<ChessState>((set, get) => ({
           autoClaimSpecialDraws: currentGame.autoClaimSpecialDraws,
           gotUndoRequest: currentGame.gotUndoRequest,
           sentUndoRequest: currentGame.sentUndoRequest,
-          info: currentGame.info
+          info: currentGame.info,
+          lastMove: currentGame.lastMove
         }
 
         set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
@@ -194,7 +198,8 @@ const useChessStore = create<ChessState>((set, get) => ({
           autoClaimSpecialDraws: setting,
           gotUndoRequest: currentGame.gotUndoRequest,
           sentUndoRequest: currentGame.sentUndoRequest,
-          info: currentGame.info
+          info: currentGame.info,
+          lastMove: currentGame.lastMove
         }
 
         set(state => ({ activeGames: state.activeGames.set(gameID, updatedGame) }))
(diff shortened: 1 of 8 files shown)


src/urbit/sur/chess.hoon
```diff
@@ -257,7 +257,7 @@
       [%challenge-received who=ship challenge=chess-challenge]
       [%challenge-resolved who=ship]
       [%challenge-replied who=ship]
-      [%position game-id=@dau position=@t san=@t special-draw-available=?]
+      [%position game-id=@dau position=@t san=@t special-draw-available=? last-move=[@t @t]]
```
ashelkovnykov: The ordering, naming, and syntax is starting to get unwieldy. Let's reset:
```
+$  chess-update
  $%  [%challenge-sent who=ship challenge=chess-challenge]
      [%challenge-received who=ship challenge=chess-challenge]
      [%challenge-resolved who=ship]
      [%challenge-replied who=ship]
      [%draw-offer game-id=@dau]
      [%draw-declined game-id=@dau]
      [%undo-declined game-id=@dau]
      [%undo-accepted game-id=@dau position=@t undo-moves=@ta]
      [%undo-request game-id=@dau]
      [%result game-id=@dau result=chess-result]
      [%special-draw-preference game-id=@dau setting=?]
  ::
      $:  %position
          game-id=@dau
          move=(pair @t @t)
          position=@t
          san=@t
          special-draw-available=?
  ==  ==
```

src/urbit/mar/chess/update.hoon
```diff
@@ -45,7 +45,7 @@
         %-  pairs:enjs
         :~  ['chessUpdate' [%s 'position']]
             ['gameID' [%s (scot %da game-id.upd)]]
-            ['move' (pairs:enjs ~[['san' [%s san.upd]] ['fen' [%s position.upd]]])]
+            ['move' (pairs:enjs ~[['san' [%s san.upd]] ['fen' [%s position.upd]] ['lastMove' [%a ~[[%s -.last-move.upd] [%s +.last-move.upd]]]]])]
```
ashelkovnykov: You're having trouble with your JSON parsing because you're attempting to include the squares of the move in the JSON object corresponding to `Move` in `frontend/ts/types/urbitChess.tx`, but you haven't extended that type to receive them.

You'll need to update the `Move` type to accept the squares, but I would handle this differently than you currently are: I'd include the squares in the `Move`, and since the `GameInfo` type already has a list of moves, you can just pull the data out of the most recent move in there. I would also recommend not treating the two moves as an `Array` until they need to be passed into Chessground - it will make dealing with them much easier.

```
%position
  %-  pairs:enjs
  :~  ['chessUpdate' [%s 'position']]
      ['specialDrawAvailable' [%b special-draw-available.upd]]
  ::
      :-  'move'
      %-  pairs:enjs
      :~  ['start' [%s p.move.upd]]
          ['end' [%s q.move.upd]]
          ['san' [%s san.upd]]
          ['fen' [%s position.upd]]
  ==  ==
```

src/frontend/ts/types/urbitChess.ts
```diff
@@ -1,3 +1,5 @@
+import * as cg from 'chessground/types'
+
```
ashelkovnykov: We only need `Key`:
```
import { Key as CgKey } from 'chessground/types'
```

src/frontend/ts/types/urbitChess.ts
```diff
@@ -103,7 +105,8 @@ export type ActiveGameInfo = {
   autoClaimSpecialDraws: boolean,
   gotUndoRequest: boolean,
   sentUndoRequest: boolean,
-  info: GameInfo
+  info: GameInfo,
+  lastMove: Array<cg.Key> | null
```
ashelkovnykov: See above comment. This change and the other one in this file can be reverted. Instead, `Move` should be modified to look like this:
```
export type Move = {
  start: CgKey
  end: CgKey
  san: SAN
  fen: FENPosition
}
```
**PS**:`from` and `to` are probably better names, but I don't want to edit my existing comments so I'm adding this thought as an addendum.

src/frontend/ts/state/chessStore.ts
```diff
@@ -65,7 +65,8 @@ const useChessStore = create<ChessState>((set, get) => ({
       autoClaimSpecialDraws: false,
       gotUndoRequest: false,
       sentUndoRequest: false,
-      info: data
+      info: data,
+      lastMove: null
```
ashelkovnykov: See above comments. All of the changes in this file can be reverted.

src/frontend/tsx/Chessboard.tsx
```diff
@@ -89,6 +89,10 @@ export function Chessboard () {
 
     if (displayGame !== null) {
       chess.load(displayGame.position)
+      const config: CgConfig = {
+        lastMove: displayGame.lastMove
+      }
+      api?.set(config)
```
ashelkovnykov: This isn't the correct place. You'll need to add a `lastMove` setting to `updateBoard`:
```
lastMove: (displayGame == null || displayGame.info.moves == null)
  ? null
  : [
    displayGame.info.moves[displayIndex].start,
    displayGame.info.moves[displayIndex].end
  ],
```

src/urbit/app/chess.hoon
```diff
@@ -1201,4 +1202,24 @@
    |=  position=chess-position
    ^-  ?
    (gte ply-50-move-rule.position 100)
+++  get-squares
+  |=  [move=chess-move player=chess-side]
+  ^-  [@t @t]
```
ashelkovnykov: See comment in `sur/chess`. Should change this to `(pair @t @t)`.

We may also want to give this type a name later. It's hardly the only one, though. Would you kindly leave an `XX` comment reminding us to do so later.

src/urbit/app/chess.hoon
```diff
@@ -1201,4 +1202,24 @@
...
+            %white  ['e1' 'c1']
+            %black  ['e8' 'c8']
+          ==
+        %kingside
+          ?-  player
+            %white  ['e1' 'g1']
+            %black  ['e8' 'g8']
+          ==
+      ==
+  ==
```
ashelkovnykov: Style:
```
?-  -.move
    %move
  :-  (cat 3 -.from.move (scot %ud +.from.move))
  (cat 3 -.to.move (scot %ud +.to.move))
::
    %castle
  ?-  +.move
      %queenside
    ?-  player
      %white  ['e1' 'c1']
      %black  ['e8' 'c8']
    ==
  ::
      %kingside
    ?-  player
      %white  ['e1' 'g1']
      %black  ['e8' 'g8']
    ==
  ==
==
```
I think I'm correct in not ending with all of the `==`s on one line - I think that's a "kingside" convention whereas the above is "queenside".

src/urbit/app/chess.hoon
```diff
@@ -1125,12 +1125,13 @@
         (check-50-move-rule u.new-position)
     ==
   =/  special-draw-claim  &(special-draw-available auto-claim-special-draws.game-state)
+  ~&  (get-squares [move player-to-move.position.game-state])
```
ashelkovnykov: Don't forget to remove this once you're satisfied with how everything is working

src/urbit/app/chess.hoon
```diff
@@ -1125,12 +1125,13 @@
...
   =/  position-update-card
     :*  %give
         %fact
         ~[/game/(scot %da game-id.game.game-state)/updates]
         %chess-update
-        !>([%position game-id.game.game-state (position-to-fen u.new-position) san special-draw-available])
+        !>([%position game-id.game.game-state (position-to-fen u.new-position) san special-draw-available (get-squares [move player-to-move.position.game-state])])
```
ashelkovnykov: Also getting unwieldy:
```
=/  position-update-card
:*  %give
    %fact
    ~[/game/(scot %da game-id.game.game-state)/updates]
    %chess-update
    !>  :*  %position
            game-id.game.game-state
            (get-squares move player-to-move.position.game-state)
            (position-to-fen u.new-position)
            san
            special-draw-available
==      ==
```

src/urbit/app/chess.hoon
```diff
@@ -1201,4 +1202,24 @@
    |=  position=chess-position
    ^-  ?
    (gte ply-50-move-rule.position 100)
+++  get-squares
```
ashelkovnykov: This should probably be a library function

src/urbit/app/chess.hoon
```diff
@@ -1125,12 +1125,13 @@
         (check-50-move-rule u.new-position)
```
ashelkovnykov: Don't forget that you'll also need to update the `%position` cards sent on line `818`. It's possible to do so without having to touch the type of `moves.game.game-state`, but you'll need to use `spin` (or possibly `spun`) instead of `turn` (let me know if that's not enough of a directional hint).

ashelkovnykov (changes requested):
I agree with your appraisal that there's no way to fix this bug other than storing the previous moves in a format that Chessground can use, and therefore this information should come from the ultimate arbiter of truth: the backend.

I really dislike this necessity, though, because this is a case of the frontend driving what information is being served by the backend. Who knows - maybe it's not so bad, though, and any frontend visualization scheme would need this information.

**Rant:** This is all because Chessground isn't React and we're trying to use React. There's a guy who made a ["React version of Chessground"](https://github.com/react-chess/chessground), but don't be fooled - I tried using it and it was an even bigger nightmare. It hides so much of Chessground's native capabilities that it becomes an altogether different hassle.

src/frontend/ts/types/urbitChess.ts
```diff
@@ -80,6 +82,8 @@ export type SAN = string
 export type FENPosition = string
 
 export type Move = {
+  start: CgKey
+  end: CgKey
```
ashelkovnykov: Name mismatch: these should be "from" and "to"

src/frontend/tsx/Chessboard.tsx
```diff
@@ -196,6 +196,12 @@ export function Chessboard () {
       fen: (displayIndex == null || displayGame.info.moves == null)
```
ashelkovnykov, line 191: Would you kindly please delete `boardTitle` (lines `60` - `64` inclusive) while you're making these changes.

src/frontend/tsx/Chessboard.tsx
```diff
@@ -196,6 +196,12 @@ export function Chessboard () {
...
         ? chess.fen()
         : displayGame.info.moves[displayIndex].fen,
+      lastMove: (displayGame == null || displayGame.info.moves == null)
+        ? null
+        : [
+          displayGame.info.moves[displayIndex].from,
+          displayGame.info.moves[displayIndex].to
+        ],
```
ashelkovnykov: There's a slight bug here. What we really want is:
- If the display game is `null` (meaning we're looking at the practice board)
    - stored prev move for practice board
- If the display game is not `null`, but there's no moves yet
    - `null`
- If the display game is not `null`, and there are moves, but `displayIndex` is `null`
    - final element in list of moves
- If the display game is not `null`, and there are moves, and `displayIndex` is not `null`
    - `displayIndex`

However, since we don't have a way to track the previous move on the practice board (and it would be surprisingly annoying to introduce it), that case should also be `null`.

src/frontend/ts/types/urbitChess.ts
```diff
@@ -1,3 +1,5 @@
+import { Key as CgKey } from 'chessground/types'
```
ashelkovnykov, line 1: Would you kindly please fix the import in `ts/types/chessground.ts` in a similar way

ashelkovnykov (changes requested):
Almost there

ashelkovnykov (approved):
LGTM. If you've verified it's all working with a fake ship, then squash the PR into one commit and feel free to merge.

src/frontend/tsx/Chessboard.tsx
```diff
@@ -196,6 +191,17 @@ export function Chessboard () {
       fen: (displayIndex == null || displayGame.info.moves == null)
         ? chess.fen()
         : displayGame.info.moves[displayIndex].fen,
+      lastMove: (displayGame == null || displayGame.info.moves == null)
+        ? null
+        : (displayIndex == null)
+          ? [
+            displayGame.info.moves[displayGame.info.moves.length - 1].from,
```
rovmug-ticfyn, line 198: These lines appear to be causing a bug on a fake ship. When selecting an active game, the screen goes blank and the JS console gives me the error: "Uncaught TypeError: t.info.moves[(t.info.moves.length - 1)] is undefined".

This case of no moves yet in a game should have been caught by line 194 checking whether `displayGame.info.moves` is `null`.

bonbud-macryg: I've had the "screen goes blank" bug a few times before and in my experience it's always because the frontend is trying to render something that's `undefined` or `null`.

src/urbit/app/chess.hoon
```diff
@@ -1126,12 +1135,17 @@
...
-        %chess-update
-        !>([%position game-id.game.game-state (position-to-fen u.new-position) san special-draw-available])
-    ==
+  :*  %give
+      %fact
+      ~[/game/(scot %da game-id.game.game-state)/updates]
+      %chess-update
+      !>  :*  %position
+              game-id.game.game-state
+              (get-squares move player-to-move.position.game-state)
```
ashelkovnykov:
```suggestion
              (get-squares move player-to-move.u.new-position)
```

src/frontend/tsx/Chessboard.tsx
```diff
@@ -196,6 +191,17 @@ export function Chessboard () {
       fen: (displayIndex == null || displayGame.info.moves == null)
         ? chess.fen()
         : displayGame.info.moves[displayIndex].fen,
+      lastMove: (displayGame == null || displayGame.info.moves == null)
```
ashelkovnykov:
```suggestion
      lastMove: (displayGame == null || displayGame.info.moves == null || (displayGame.info.moves.length === 0))
```

ashelkovnykov (changes requested):
Works after these two changes

ashelkovnykov (approved):
Works correctly, looks really nice now

PR #123, opened 2023-04-01
Add log messages for bad gameID in frontend

ashelkovnykov:


src/frontend/ts/state/chessStore.ts
```diff
@@ -5,6 +5,10 @@ import { Update, Ship, GameID, SAN, FENPosition, Move, GameInfo, ActiveGameInfo,
 import { findFriends } from '../helpers/urbitChess'
 import ChessState from './chessState'
 
+const badGameId = (gameID: GameID) => {
+  console.log('received bad gameId: ' + gameID)
+}
+
```
bonbud-macryg, line 13: Is "bad gameId" the most helpful message for this, or should it specifically mention that this is for a gameId that's no longer active? A game you don't have? Unless someone's screwing around in their code, the circumstance where this is going to be necessary is something like: you've abandoned a game, but your opponent didn't receive word of this for some reason.

bonbud-macryg (approved):
I was tinkering with some code on two fakeships to try and test this but it's not worth contriving a situation where this message would be necessary. I just have the one comment about the wording of the message but other than that: LGTM.

ashelkovnykov:
@bonbud-macryg I also just tested it to make sure that it didn't break the existing functionality, but didn't bother to contrive a situation where it actually happened.

src/frontend/ts/state/chessStore.ts
ashelkovnykov: It's not the most helpful message - ideally, we'd also be logging which function was called with the bad ID and checking whether the ID is in the archive. I just wanted to get ***some*** kind of safety check out the door.

You're right, though, that at the very least we should add what I just wrote above as a comment.

bonbud-macryg (approved):

PR #124, opened 2023-04-01
Fix draw/undo requests

### LINKED ISSUE #114: Pop-ups don't reappear
If a pop-up for a draw offer or undo is closed without resolution, it cannot later be reopened. I think a good solution is to add a button to the `GamePanel` which become enabled for the player receiving the draw offer/undo request can cancel the pop-up (this is how lichess does it, without any pop-ups).

ashelkovnykov:
Resolves #114. The "send draw offer" button will become an "accept draw offer" button whenever a draw offer is received. The "request to undo move" button will become an "accept undo request" button whenever an undo request is received. The pop-ups still appear, but if they're closed, then the user still has access to the new buttons.


bonbud-macryg (approved):
Big improvement; LGTM.

PR #111, opened 2023-02-01
Release Candidate: 0.9.2

ashelkovnykov:
Atypical release candidate (since this is the first one); all changes going in were force-pushed earlier when the history of `master` was rewritten.

**Status**: Tested successfully

PR #126, opened 2023-04-01
Update package.json and build.sh

ashelkovnykov:
- Missed the version in `package.json` when bumping to `0.9.3`
- Added `DOCKER_BUILDKIT` workaround to `build.sh` so that build works for MacOS users hitting [this issue](https://github.com/moby/buildkit/issues/1271)

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -136,7 +136,10 @@ echo "[%zuse $KELVIN]" > $DESK_DIR/sys.kelvin
 
 # Build frontend
 if [ $DOCKER -eq 1 ]; then
-  sudo docker build --tag ${DOCKER_IMAGE}:${VERSION_FULL} .
+  # Need to use legacy builder ( DOCKER_BUILDKIT=0) so that MacOS builds work until this issue is resolved:
+  #   https://github.com/moby/buildkit/issues/1271
+  # sudo docker build --tag ${DOCKER_IMAGE}:${VERSION_FULL} .
+  sudo DOCKER_BUILDKIT=0 docker build --tag ${DOCKER_IMAGE}:${VERSION_FULL} .
   sudo docker run --rm -v ${FRONTEND_DIR}:/app/output/ ${DOCKER_IMAGE}:${VERSION_FULL}
 
   # Copy additional src files for frontend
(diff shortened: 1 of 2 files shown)


bonbud-macryg (approved):
LGTM. Built with webpack with no issue.

Built with Docker and had the same issue I had with #111, requiring this change to `build.sh`. (Old version of the amended line is commented out.)

```
if [ $DOCKER -eq 1 ]; then
  sudo DOCKER_BUILDKIT=0 docker build --tag ${DOCKER_IMAGE}:${VERSION_FULL} .
  # sudo docker build --tag ${DOCKER_IMAGE}:${VERSION_FULL} .
  sudo docker run --rm -v ${FRONTEND_DIR}:/app/output/ ${DOCKER_IMAGE}:${VERSION_FULL}

  # Copy additional src files for frontend
  sudo chown -R ${USER}:${USER} ${FRONTEND_DIR}
else
  (cd "$ROOT_DIR/src/frontend"; npm run build-no-docker)
fi
```
Is it worth adding that `DOCKER_BUILDKIT=0` into `build.sh` in a separate PR? I'm not sure what's causing it on my machine or how common it would be for other macOS users.

ashelkovnykov:
@bonbud-macryg Good idea - I'll add it as a separate commit in this PR. We should absolutely make sure that we have a working build path for MacOS users.

bonbud-macryg:
@ashelkovnykov I'm not sure it's specifically a macOS issue on my end, but I'm the only one on the Chess team who's had this issue so it's my first suspect. I'd never had the issue till #111 and I hadn't changed anything with Docker. In any case `DOCKER_BUILDKIT=0` eliminates the error and afaict there's no cost to having it there, so `¯\_(ツ)_/¯`

ashelkovnykov:
@bonbud-macryg It's specifically a MacOS issue, but not just on your end. Like you say, there's no cost to disabling the Docker Buildkit - it's just the package they use for building an image from a Dockerfile in the latest version of Docker. If you run with it disabled, it gives a warning that running without Buildkit is deprecated, so the next version of Docker will remove the old build system. However, I'd hope that by then they fix an issue blocking building on MacOS...

PR #127, opened 2023-04-01
Fix minor visual bug

ashelkovnykov:
Before this fix, any time an update would come through while a user was browsing an old board state it would reset the board. Now, the board is only reset to the current state when an undo request is accepted or when a new move is made.

### DIFF
--- a/src/frontend/ts/state/chessStore.ts
+++ b/src/frontend/ts/state/chessStore.ts
@@ -81,7 +81,7 @@ const useChessStore = create<ChessState>((set, get) => ({
   receiveUpdate: (data: ChessUpdate) => {
     const updateDisplayGame = (updatedGame: ActiveGameInfo) => {
       if ((get().displayGame !== null) && (updatedGame.info.gameID === get().displayGame.info.gameID)) {
-        set({ displayGame: updatedGame, displayIndex: null })
+        set({ displayGame: updatedGame })
       }
     }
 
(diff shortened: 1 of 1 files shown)


bonbud-macryg:
I think this behaviour was actually a deliberate decision of mine but it's not ideal. In this final version there should probably be some kind of non-invasive alert or visual cue that a new move was made while you were viewing previous states.

ashelkovnykov:
@bonbud-macryg I agree - I was thinking of heading into v1 like this and waiting for the community to share feedback, on whether this is annoying or not. Also wanted to do a live sit-down with some of the Tlon designers and get their feedback.

bonbud-macryg (commented):
This works as intended, though I'm seeing the bug #129 quite badly on this commit having not seen it in months. I don't *suspect* that the changes here are causing it, but I might have to do a git-bisect when I have time to confirm.

bonbud-macryg:
Worth noting I played a 15-move game on the same ships while testing #124 with no issues.

ashelkovnykov:
#118 Had a bug in it which would cause this, though it was addressed and tested successfully. I tested both this commit and #118 for several moves without issue. Standby, I'll rebase onto develop and re-test both develop and this PR.

ashelkovnykov:
Just added a comment to #129 - relevant bit is that I tested this PR on completely new fake ships and didn't encounter any issues.

bonbud-macryg (approved):

bonbud-macryg:
@ashelkovnykov Comment in #129 sounds right; I'll approve this now and test #129 later to confirm / get the process down.

PR #134, opened 2023-05-02
Display material difference

### LINKED ISSUE #11: Show material difference while playing
It would be nice to see what the material difference in pieces and value is between players during a match. Should be pretty easy to compute from the FEN.

Frontend work only.

rovmug-ticfyn:
This PR will resolve #11 by displaying the material difference. Initial commit just displays a number for the material advantage, then display of pieces will be added.

### DIFF
--- a/src/frontend/tsx/GamePanel.tsx
+++ b/src/frontend/tsx/GamePanel.tsx
@@ -79,6 +79,57 @@ export function GamePanel () {
     }
   }
 
+  const materialDifference = (fen: string) => {
+    let board = fen.split(' ')[0]
+    if (displayIndex !== null) {
+      board = displayGame.info.moves[displayIndex].fen.split(' ')[0]
+    }
+    const whitePieces: { [key: string]: number } = { 'q': 0, 'r': 0, 'b': 0, 'n': 0, 'p': 0 }
+    const blackPieces: { [key: string]: number } = { 'q': 0, 'r': 0, 'b': 0, 'n': 0, 'p': 0 }
+    let whiteDisplay = ''
+    let blackDisplay = ''
+
+    for (let i = 0; i < board.length; i++) {
+      const piece = board.charAt(i)
+      if (piece === '/') {
+        continue
+      }
+      if (isNaN(parseInt(piece))) {
+        if (piece === piece.toUpperCase()) {
+          whitePieces[piece.toLowerCase()] += 1
+        } else {
+          blackPieces[piece] += 1
+        }
+      } else {
+        continue
+      }
+    }
+
+    for (let key in whitePieces) {
+      if (whitePieces[key] > blackPieces[key]) {
+        whiteDisplay += Array(whitePieces[key] - blackPieces[key]).fill(CHESS.pieceIconsBlack[key]).join('')
+      } else if (blackPieces[key] > whitePieces[key]) {
+        blackDisplay += Array(blackPieces[key] - whitePieces[key]).fill(CHESS.pieceIconsWhite[key]).join('')
+      }
+    }
+
+    const materialDifference = Object.keys(CHESS.pieceValues).reduce(
+      (diff, key) => diff + (whitePieces[key] - blackPieces[key]) * CHESS.pieceValues[key],
+      0
+    )
+
+    if (materialDifference > 0) {
+      whiteDisplay += '+' + materialDifference
+    } else if (materialDifference < 0) {
+      blackDisplay += '+' + Math.abs(materialDifference)
+    }
+
+    return {
+      white: whiteDisplay,
+      black: blackDisplay
+    }
+  }
+
   //
   // HTML element generation functions
   //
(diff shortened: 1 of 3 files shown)


src/frontend/tsx/GamePanel.tsx
```diff
@@ -66,6 +66,51 @@ export function GamePanel () {
...
+    let whiteValue = 0
+    let blackValue = 0
+    const captured = []
+
+    for (let i = 0; i < board.length; i++) {
+      const piece = board[i]
+      if (piece === '/') {
+        continue
+      }
+      const isWhite = piece === piece.toUpperCase();
```
ashelkovnykov: This is going to always return `false` for numeric characters. You need to do a check for `!isNaN(Number(piece))` before doing the case check.

src/frontend/tsx/GamePanel.tsx
```diff
@@ -142,6 +187,14 @@ export function GamePanel () {
   return (
     <div className='game-panel-container col' style={{ display: ((displayGame !== null) ? 'flex' : ' none') }}>
       <div className="game-panel col">
+        <div id="opp-captured" className={'captured row' + (hasGame ? '' : ' hidden')}>
+          <p>
+            { (displayGame.info.white !== window.ship)
+            ? (materialDifference(displayGame.position, 'black'))
+            : (materialDifference(displayGame.position, 'white'))
```
ashelkovnykov: I think you accidentally made a double-negative here, by swapping both `===` to `!==` and `white` to `black`: I'm pretty sure both of these displays are doing to show the same side's material difference

src/frontend/tsx/GamePanel.tsx
```diff
@@ -66,6 +66,51 @@ export function GamePanel () {
...
+      const isWhite = piece === piece.toUpperCase();
+      const pieceValue = CHESS.pieceValues[piece.toLowerCase()];
+      if (isWhite) {
+        whiteValue += pieceValue;
+      } else {
+        blackValue += pieceValue;
+      }
+      if (isNaN(Number(piece))) {
+        if ((side === 'white' && isWhite) || (side === 'black' && !isWhite)) {
+          captured.push(piece.toLowerCase())
```
ashelkovnykov: I'm not really sure how this `captured` system is supposed to work. I left my thoughts for the general algorithm I'd use for calculating material difference in the main comment, but please let me know what you're thinking here if you plan to keep going in this direction.

ashelkovnykov (changes requested):
My general advice for computing material difference would be to only scan through the board once. Have the material difference calculation function return the difference for both sides as a map (e.g. `{'white': ..., 'black': ...}`). Alternatively, you could use the ship names as the map keys, if that makes accessing the result data easier.

Here's the general material difference computation algorithm that I would go with:
1. Iterate over the board FEN, counting the number of each type of piece for each side
2. Iterate over the list of each piece type, `P`
3. For `P`, determine who has more pieces: white, black, or neither
4. If neither, continue
5. If white (black), then record how many more of piece `P` white (black) has than black (white) in `whiteMaterialRenderMap` (`blackMaterialRenderMap`) and add the associated number of points to `whiteMaterialValue` (`blackMaterialValue`)
6. After you're done, you should have a rendering map for each side that looks something like: `{'p': 3, 'b': 2, 'k': 1, 'r': 0, 'q': 0}` and a material value count. Note that this map records how many pieces to render to the screen in the "captured" zone of a particular side. Note also that it's technically not perfect as it will presume a lost piece, if for example both sides still have both rooks, but one player manages to promote a pawn to a rook. However, these cases are so rare that it's not important, and the graphics will still correctly convey the valuable information that "the opponent is up a rook".
7. When rendering the "captured" zone, iterate over each piece type and draw as many of the piece as exists in that side's `RenderMap`
8. After rendering the pieces, render the material difference for whichever player has more material

rovmug-ticfyn:
Your approach to the material difference algorithm makes sense, so I'll begin making those changes. In terms of efficiency, should the material difference be stored in frontend state and only recomputed when a piece is captured?

ashelkovnykov:
@rovmug-ticfyn That's a good question. Really, it should be stored in the backend state:

It would be easier to compute the material difference in the backend as the game went along: when there's a capture, record the captured piece and increment / decrement the material value of white / black. This is the kind of info any chess UI would be interested in rendering, so it makes sense to store it in the backend.

However, it's annoying to deal with the marks, the undo logic, the JS objects, etc. when adding a data field to the backend that needs to be sent to the frontend. Computing the material difference in the frontend as you are now (each iteration, just before rendering, without storing it) is easy and self-contained and gets us a working material difference display.

Caching the material difference in a `map: FEN -> result` would certainly improve the efficiency somewhat, but ultimately it's a half-measure since we should pull this into the backend.

That's all context to say: I leave it up to you. I think any of the above solutions are acceptable, since they all get us a working material difference display and the sub-optimal ones are quick & easy to implement, and easily replaceable.

ashelkovnykov (commented):
Code looks good, but material changes aren't rendering. For some reason, both `whiteDisplay` and `blackDisplay` are always rendered as empty strings. Make sure to rebase your branch onto the latest `develop` for testing.

rovmug-ticfyn:
Material difference is now rendering, but I need to continue debugging some incorrect behavior.

ashelkovnykov (commented):
Looks and works great. Only remaining issue: the material difference doesn't update when looking at past moves.

ashelkovnykov (approved):
:+1:

PR #135, opened 2023-05-06
Move Arvo dependencies into repo

ashelkovnykov:
Move `base-dev` + `desk-dev` dependency files into the repo


bonbud-macryg (approved):
LGTM; built with Webpack and Docker.

PR #139, opened 2023-05-12
Release 0.9.4 for %413

ashelkovnykov:
Tested locally


bonbud-macryg (approved):
LGTM. Built with webpack and Docker and made a few moves with the practice board, not tested beyond that.

PR #143, opened 2023-08-07
Add -k option to build script

bonbud-macryg:
The %chess desk's kelvin version decreases regularly, but without regular maintenance our development fakeships lag behind. When we have a %chess kelvin lower than our fakeship's, we can either…
* Download the latest urbit binaries and build a new fakeship.
* Manually change the kelvin in `build.sh` and undo the change once we're ready to send in a PR.

Neither of these are ideal. To avoid this choice, this PR adds a `-k` option to the `build.sh` script that lets us specify the kelvin version we want to build with.

As far as I can tell most kelvin changes don't affect %chess functionality - manually changing the kelvin in `build.sh` has never caused an issue for me - so in almost all cases using this option is just a formality to make the compiler happy.

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -22,7 +22,7 @@ usage() {
   fi
 
   echo -e ""
-  echo -e "Usage:\t$SCRIPT_NAME [-h] [-n] [-s SHIP_NAME] [-u URL]"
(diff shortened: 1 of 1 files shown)


ashelkovnykov (approved):
LGTM

PR #147, opened 2023-08-25
Display material difference

rovmug-ticfyn:
This is my first pass at calculating material difference on the backend. Still need to go through and edit all places where data needs to be sent to the frontend—along with marks—and add a frontend display.

### DIFF
--- a/src/backend/chess/app/chess.hoon
+++ b/src/backend/chess/app/chess.hoon
@@ -18,6 +18,9 @@
       position=chess-position
       fen-repetition=(map @t @ud)
       special-draw-available=?
+      white-has-captured=(map chess-piece @ud)
+      black-has-captured=(map chess-piece @ud)
+      material-advantage=[@ud $?(chess-side %none)]
       auto-claim-special-draws=?
       sent-draw-offer=?
       got-draw-offer=?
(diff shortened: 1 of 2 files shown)


ashelkovnykov (commented):
Surprisingly, I prefer your first version which computed the material difference on the frontend. I'm finding this one harder to follow, and it seems like the algorithm was a lot more compact on the frontend.

If we already had multiple frontends, it might be a nice optimization to track this info for a game and let the frontends scry it. For now, I think it'd be faster to fix up the bugs in the algorithm for your initial version and come back to this at a later date.

If you insist on going with a backend version:
- It'd be simpler to have `apply-move` return a `(unit chess-piece)` for a move
  - It could do this by caching the piece at the destination square and returning it if the move is valid
- Add the captured piece to a set of pieces for the appropriate side
- At the same time as the above, increment a counter containing the value of the captured pieces per side
- For valid promotion, additionally add a pawn / the value of a pawn, but subtract a `X` / the value of an `X`
- Don't proactively pass this info along with the game state; let the frontends scry for it
- I'm not sure how to handle undo logic - it might be so troublesome that you would need to abort the above, and just do what you did earlier in the frontend but in the backend: just scan the final position after any move and get the data from that

rovmug-ticfyn:
Yes, it was certainly more compact on the frontend. The undo logic added even more complexity to this because I decided to just iterate through the position on an undo and rebuild the map of captured pieces.

I think you're right that the frontend version will overall be a lower lift if I can get that algorithm working.

PR #148, opened 2023-09-07
Show win/loss record

### LINKED ISSUE #17: Show current record against opponent
After #6 is implemented, it would be trivial to quickly lookup a player's record against their opponent and add it to a badge somewhere on-screen.

Frontend work only.

bonbud-macryg:
Resolves #17.

### DIFF
--- a/src/backend/chess/mar/chess/game/active.hoon
+++ b/src/backend/chess/mar/chess/game/active.hoon
@@ -1,18 +1,9 @@
-<<<<<<< HEAD
 /-  *chess
 =,  format
 |_  game=chess-game
 ++  grab
   |%
   ++  noun  chess-game
-=======
-/+  chess
-=,  format
-|_  game=chess-game:chess
-++  grab
-  |%
-  ++  noun  chess-game:chess
->>>>>>> 6f463aa (add functionality for browsing completed games)
   --
 ++  grow
   |%
@@ -26,11 +17,7 @@
         ['archived' [%b %.n]]
         ::  default values. updated thru [%game @ta %updates ~] sub.
         ['moves' [%a ~]]
-<<<<<<< HEAD
         ['position' [%s *chess-fen]]
-=======
-        ['position' [%s '']]
->>>>>>> 6f463aa (add functionality for browsing completed games)
         ['gotDrawOffer' [%b %.n]]
         ['sentDrawOffer' [%b %.n]]
         ['drawClaimAvailable' [%b %.n]]
--- a/src/backend/chess/mar/chess/game/active.hoon
+++ b/src/backend/chess/mar/chess/game/active.hoon
@@ -1,9 +1,18 @@
+<<<<<<< HEAD
 /-  *chess
 =,  format
 |_  game=chess-game
 ++  grab
   |%
   ++  noun  chess-game
+=======
+/+  chess
+=,  format
+|_  game=chess-game:chess
+++  grab
+  |%
+  ++  noun  chess-game:chess
+>>>>>>> 6f463aa (add functionality for browsing completed games)
   --
 ++  grow
   |%
@@ -17,7 +26,11 @@
         ['archived' [%b %.n]]
         ::  default values. updated thru [%game @ta %updates ~] sub.
         ['moves' [%a ~]]
+<<<<<<< HEAD
         ['position' [%s *chess-fen]]
+=======
+        ['position' [%s '']]
+>>>>>>> 6f463aa (add functionality for browsing completed games)
         ['gotDrawOffer' [%b %.n]]
         ['sentDrawOffer' [%b %.n]]
         ['drawClaimAvailable' [%b %.n]]
--- a/src/frontend/ts/helpers/chess.ts
+++ b/src/frontend/ts/helpers/chess.ts
@@ -36,13 +36,16 @@ export function getTally (ship: Ship): String {
 
   let tally: string;
 
+  console.log('getTally tallies', tallies)
   if (!tallies.has(ship)) {
     tally = '0 - 0';
   } else {
       const results = tallies.get(ship)
-      const ourWins = results.losses + Math.floor(results.draws / 2)
-      const oppWins = results.wins + Math.floor(results.draws / 2)
-      const draws = (results.draws % 2) === 1 ? "½" : "";
+      console.log('getTally results:', results);
+      const ourWins = results.losses
+      const oppWins = results.wins
+      const draws = (results.draws % 2) === 1 ? '½' : '';
+      console.log('getTally draws:', draws);
 
       tally = `${ourWins}${draws} - ${oppWins}${draws}`
   }
--- a/src/frontend/ts/helpers/chess.ts
+++ b/src/frontend/ts/helpers/chess.ts
@@ -1,5 +1,6 @@
 import { ChessInstance, Square, SQUARES, WHITE, PAWN } from 'chess.js'
 import { Ship, Results } from '../types/urbitChess'
+import useChessStore from '..//state/chessStore'
 
 /**
  * Determine if a chess move is a pawn promotion
@@ -30,7 +31,13 @@ export function getChessDests (chess: ChessInstance): Map<string, string[]> {
   return dests
 }
 
-export function getTally (ship: Ship, results: Results): String {
+export function getTally (ship: Ship): String {
+  const { tallies } = useChessStore()
+
+  const outputTally: String = '420 - 69'
+
+  return outputTally
+
   // console.log('ship')
   // console.log(ship)
   // console.log('results')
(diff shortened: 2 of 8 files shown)


src/backend/chess/mar/chess/game/active.hoon
ashelkovnykov, line 1: Rebase errors in this file

src/frontend/tsx/Challenges.tsx
```diff
@@ -201,10 +206,34 @@ export function Challenges () {
...
           </div>
+          <div
+            className="new-opp-tally-container"
+            style={
+              newOpp === ''
+              ? {display: 'none'}
+              : !ob.isValidPatp(`${newOpp}`)
+              ? {display: 'none'}
+              : {display: 'block'}
+            }
```
ashelkovnykov: I don't think the linter throws errors about this, but the indentation should be:
```
newOpp === ''
  ? {display: 'none'}
  : !ob.isValidPatp(`${newOpp}`)
    ? {display: 'none'}
    : {display: 'block'}
```

src/frontend/tsx/Challenges.tsx
```diff
@@ -178,19 +183,19 @@ export function Challenges () {
                   <div className='row'>
                     <div className='col'>
                       <p className='friend'>~{friend}</p>
-                      {/* XX: win/loss history with this friend */}
-                      <p className='score'>0-0</p>
+                      <p className='score'>{( tallies.get(`~${friend}`) === undefined ? '0 - 0' : tallies.get(`~${friend}`))}</p>
```
ashelkovnykov: Check can be simplified:
```
tallies.has(`~${friend}`) ? '0 - 0' : tallies.get(`~${friend}`)
```

src/frontend/tsx/Challenges.tsx
```diff
@@ -201,10 +206,34 @@ export function Challenges () {
...
+          >
+            <p className="new-opp-tally">
+                { newOpp === ''
+                  ? ''
+                  : !ob.isValidPatp(`${newOpp}`)
+                  ? ''
+                  : (tallies.get(`${newOpp}`) === undefined
+                  ? '0 - 0'
+                  : tallies.get(`${newOpp}`))
+                }
```
ashelkovnykov: Since the checks for the `<div>` visibility above are already hiding / showing the text as necessary, we only need to check for whether we have a tally for the player. Combined with the above two comments, this looks like:
```
{ tallies.has(`${newOpp}`)
    ? '0 - 0'
    : tallies.get(`${newOpp}`))
}
```

src/frontend/ts/state/chessStore.ts
```diff
@@ -112,6 +114,58 @@ const useChessStore = create<ChessState>((set, get) => ({
     await get().fetchArchivedMoves(gameID)
     get().setDisplayGame(get().archivedGames.get(gameID))
   },
+  countTallies: async (archivedGames: Map<GameID, ArchivedGameInfo>) => {
+    const tallies = get().tallies
+    // initialize empty record of players' wins, losses, and draws
+    const winsLossesDraws: Map<Ship, { wins: number, losses: number, draws: number }> = new Map();
```
ashelkovnykov: `{ wins: number, losses: number, draws: number }` should have its own type (interface) in a new `chess.ts` file in `ts/types/`:
```
export interface Record {
  wins: number,
  losses: number,
  draws: number
}
```

src/frontend/ts/state/chessStore.ts
```diff
@@ -112,6 +114,58 @@ const useChessStore = create<ChessState>((set, get) => ({
     await get().fetchArchivedMoves(gameID)
     get().setDisplayGame(get().archivedGames.get(gameID))
   },
+  countTallies: async (archivedGames: Map<GameID, ArchivedGameInfo>) => {
```
ashelkovnykov: We shouldn't be recomputing every tally for every player each time we receive a new finished game. `tallies` should be a `Map<Ship, Record>`. When we receive a new archived game, we read the result and increment the appropriate field in the `Record` for the opponent `Ship`. Instead of `countTallies`, we have a helper function in `ts/helpers/chess.ts` that will convert a `Record` into a `String`. We call this function on the `.get()` output of `tallies` in `Challenges.tsx`. We don't need to track the numbers for both players because they're implicitly tracked: our wins are the opponents losses, and vice versa.

ashelkovnykov (changes requested):

src/frontend/ts/state/chessStore.ts
ashelkovnykov: I wrote a more-or-less complete version of what `receiveArchivedGame` should look like for this:
```
const tallies = get().tallies
let opponent = data.white === window.ship
  ? data.black
  : data.white
let record = tallies.has(opponent)
  ? tallies.get(opponent)
  : new Map().set(opponent, { wins: 0, losses: 0, draws: 0 });
let newRecord = null

switch (data.result) {
  case "1-0":
    if (data.white == window.ship) {
      newRecord = {
        ...record,
        wins: record.wins + 1
      }
    } else {
      newRecord = {
        ...record,
        losses: record.losses + 1
      }
    }
    break;
  case "0-1":
    if (data.white == window.ship) {
      newRecord = {
        ...record,
        losses: record.losses + 1
      }
    } else {
      newRecord = {
        ...record,
        wins: record.wins + 1
      }
    }
    break;
  case "½–½":
    newRecord = {
      ...record,
      draws: record.draws + 1
    }
    break;
  default:
    // XX: Need to print message to console and error out
}

set(state => ({
  archivedGames: state.archivedGames.set(data.gameID, data)
  tallies: state.tallies.set(opponent, newRecord)
  }))
```

bonbud-macryg:
Bug: draws aren't shown in the UI tally.

bonbud-macryg:
I think I've found the culprit, `receiveArchivedGame` is storing two versions of our (~dev) results in `tallies`. It's also not storing draws in opponents' results. Will address later.

ashelkovnykov:
@bonbud-macryg The code itself looks good. Are the `~dev` games from playing against yourself? I think that might confuse the algorithm in `chessStore`. We should probably have code to ignore computing tallies when our ship is both sides.

bonbud-macryg:
@ashelkovnykov Having a safeguard against playing yourself is a good idea, I should add that in. But all these games are between ~dev and ~ser. As of `c6ac28c` ~dev is seeing the right output but ~ser isn't seeing draws in the `tallies` results from `receiveArchivedGames`.

bonbud-macryg:
Found the source of the bug: When a game ends in a draw, the player who made the final move correctly logs this but the other player never receives the move.

This is reflected both in each player's `Results` tuple (logged in both `receiveArchivedGame` and `getTally`) and in the backend, checked via `+dbug`. The player who should have received the concluding move still has the game stored in their active games.

Fastest way to draw a game [here](https://www.chess.com/forum/view/fun-with-chess/fastest-stalemate).

bonbud-macryg:
Looks like this bug is present in `develop` as well. I remember testing this PR with no issue before your review, back when it was based on PR #109 rather than `develop`.

bonbud-macryg:
Some outstanding linter issues still to be taken care of.

src/frontend/ts/helpers/chess.ts
```diff
@@ -28,3 +30,26 @@ export function getChessDests (chess: ChessInstance): Map<string, string[]> {
...
+
+  if (ship === `~${urbit.ship}`) {
+    tally = ''
+  } else {
+    if (!tallies.has(ship)) {
+      tally = '0 - 0'
+    } else {
+      const results = tallies.get(ship)
+      const ourWins = results.losses
+      const oppWins = results.wins
```
ashelkovnykov: Weren't these `results.losses + (results.draws / 2)` before? Why was that changed to this?

ashelkovnykov (changes requested):

src/frontend/ts/helpers/chess.ts
bonbud-macryg: Yeah, that `+ Math.floor(results.draws / 2)` was first-draft GPT code carried over from the original commit that I figured was unnecessary while making revisions to this. Unnecessary code was characteristic of its output.

Here's the earlier version:

```typescript
const results = tallies.get(ship)
const ourWins = results.losses + Math.floor(results.draws / 2)
const oppWins = results.wins + Math.floor(results.draws / 2)
const draws = (results.draws % 2) === 1 ? "½" : "";
```

If an opponent's `results` were `{wins: 2, losses: 2, draws: 3}` in the `tallies` map this `getTally` function would halve `3` to get `1.5`, `floor` that to `1`, add that to both `wins` and `losses` then represent the tally as if the results had been `{wins: 3, losses: 3, draws: 3}`, outputting `'3½ - 3½'`. I think the correct tally would be `'2½ - 2½'`

Here I'm logging the input `wins.results`, `losses.results`, `draws.results` and the output `oppWins`, `ourWins`, `draws.results`. (Logged twice as ~dev and ~ser are in the same Firefox window.)

ashelkovnykov: `'3½ - 3½'` is the correct tally based on typical chess scoring

ashelkovnykov (approved):

PR #152, opened 2023-09-12
Add eslint to webpack workflow

bonbud-macryg:
The webpack workflow was supposed to free developers from the requirement to use Docker. In practice, this isn't tenable without worrying about whether the Docker workflow will be blocked by linter errors they didn't catch. They have to remember to run eslint natively, or test the build with Docker.

This PR adds eslint to the webpack workflow. By default, it copies the behaviour of the Docker workflow and refuses to build if there are errors.

This also adds an optional `-l` flag to fix errors automatically. Devs can skip hunting down and correcting dozens of lines of code, but no changes are made without their say-so.

### DIFF
--- a/bin/build.sh
+++ b/bin/build.sh
@@ -22,13 +22,14 @@ usage() {
   fi
 
   echo -e ""
-  echo -e "Usage:\t$SCRIPT_NAME [-h] [-k KELVIN] [-n] [-s SHIP_NAME] [-u URL]"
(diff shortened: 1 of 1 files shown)


bonbud-macryg:
@ashelkovnykov Would you like me to add in a case for when `DOCKER` and `LINT_FIX` are both `1`, making the `-l` flag compatible with Docker builds?

ashelkovnykov:
@bonbud-macryg No, that's fine for now. I'll create a new issue for this and assign it to you, but the other stuff for the Oct 1 deadline is more urgent.

ashelkovnykov (approved):

PR #154, opened 2023-09-14
Practice matches

### LINKED ISSUE #25: Practice matches and Official matches
There should be different game modes. Practice matches that do not get recorded and Official matches where the Win/Lose Ratio is recorded.

rovmug-ticfyn:
Resolves #25.

Currently doesn't display on the list of games that a game is practice, only on the challenges. That boolean could be added to the frontend game state if we feel it's important to display.

### DIFF
--- a/src/backend/chess/app/chess.hoon
+++ b/src/backend/chess/app/chess.hoon
@@ -24,6 +24,7 @@
       sent-undo-request=?
       got-undo-request=?
       opponent=ship
+      practice-game=?
   ==
 +$  state-1
   $:  %1
@@ -614,7 +615,7 @@
             ::  remove our challenge from challenges-sent
             challenges-sent  (~(del by challenges-sent) src.bowl)
             ::  put our new game into the map of games
-            games  (~(put by games) game-id.action [new-game *chess-position *(map @t @ud) | | | | | | src.bowl])
+            games  (~(put by games) game-id.action [new-game *chess-position *(map @t @ud) | | | | | | src.bowl practice-game.u.challenge])
           ==
         %draw-offered
           =/  game-state
@@ -847,7 +848,7 @@
                 ?:  ?|  sent-draw-offer.u.game-state
                         special-draw-available.u.game-state
                     ==
-                  (output-quip game.u.game-state(result `result.action) ~)
+                  (output-quip game.u.game-state(result `result.action) ~ practice-game.u.game-state)
                 %+  poke-nack  this
                 "{<our.bowl>} did not send draw offer for {<game-id.action>}"
               ::  is opponent resigning?
@@ -855,7 +856,7 @@
                   ?:  =(our.bowl white.game.u.game-state)
                     %'1-0'
                   %'0-1'
-                (output-quip game.u.game-state(result `result.action) ~)
+                (output-quip game.u.game-state(result `result.action) ~ practice-game.u.game-state)
               %+  poke-nack  this
               "{<our.bowl>} does not resign game {<game-id.action>}"
             ::  apply move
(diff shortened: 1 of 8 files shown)


src/backend/chess/mar/chess/update.hoon
```diff
@@ -21,13 +21,15 @@
             ['who' [%s (scot %p who.upd)]]
             ['challengerSide' [%s challenger-side.challenge.upd]]
             ['event' [%s event.challenge.upd]]
+            ['practiceGame' [%b practice-game.challenge.upd]]
```
ashelkovnykov: `'isPractice'` would be better. Typically `boolean` values have a name starting with "is" or "has" (like `isBig` or `hasData`).

src/backend/chess/mar/chess/update.hoon
```diff
@@ -21,13 +21,15 @@
...
         ==
       %challenge-received
         %-  pairs:enjs
         :~  ['chessUpdate' [%s 'challenge-received']]
             ['who' [%s (scot %p who.upd)]]
             ['challengerSide' [%s challenger-side.challenge.upd]]
             ['event' [%s event.challenge.upd]]
+            ['practiceGame' [%b practice-game.challenge.upd]]
```
ashelkovnykov: See above

src/frontend/ts/types/urbitChess.ts
```diff
@@ -122,7 +122,8 @@ export interface Move {
 export interface Challenge {
   who: Ship,
   challengerSide: Side,
-  event: string
+  event: string,
+  practiceGame: Boolean
```
ashelkovnykov: See above

src/backend/chess/app/chess.hoon
```diff
@@ -1433,10 +1439,15 @@
           =*  updated-game  game.game-state
           =.  result.updated-game  `result
           :_
+            ?:  =(& practice-game.game-state)
+              %=  this
+                ::  remove this game from our map of active games
+                games    (~(del by games) game-id)
+              ==
```
ashelkovnykov, line 1451: Gah... `+output-quip` is now uglier than I'd like, but there's really no better way to do this...

ashelkovnykov (changes requested):
> Currently doesn't display on the list of games that a game is practice, only on the challenges. That boolean could be added to the frontend game state if we feel it's important to display.

Yes, please.

There's also a bug currently: once a practice game ends, it gets added to the archive due to the code on lines `223 - 226` of `chessStore`. Then, once the user refreshes the window, it's correctly removed from the archive.

ashelkovnykov (changes requested):
What you did makes sense, but I hate it. Can you drop the commit that you added to add the info to the game card? We'll merge without it. I want to re-evaluate how best to get that info to the frontend. As far as I can tell, our options are:

1. Your solution
2. Moving 'isPractice' into `chess-game`
3. A new mark for `[chess-game isPractice=?]`

They're all stupid and ugly, so I want to compare them in the code, which is a very low-priority activity.

src/frontend/ts/state/chessStore.ts
```diff
@@ -220,10 +220,16 @@ const useChessStore = create<ChessState>((set, get) => ({
...
+        if (currentGame.isPractice === false) {
+          set(state => ({
+            activeGames: activeGames,
+            archivedGames: state.archivedGames.set(gameID, updatedGame)
+          }))
+        } else {
+          set(state => ({
+            activeGames: activeGames
+          }))
+        }
```
ashelkovnykov: Can you please comment out the lines wrapping the old `set` statement, and add a `TODO` comment about how we need a way to get practice info for games from the backend to the frontend.

`currentGame` no longer has a `practice` field after removing the commit which introduced it, meaning that we'll have to live with the bug for now.

ashelkovnykov (changes requested):

ashelkovnykov (approved):

PR #156, opened 2023-09-16
Fix bug preventing stalemates

### LINKED ISSUE #155: Stalemates don't end correctly as draws for the player receiving the move
> Found the source of the bug: When a game ends in a draw, the player who made the final move correctly logs this but the other player never receives the move.
>
> This is reflected both in each player's `Results` tuple (logged in both `receiveArchivedGame` and `getTally`) and in the backend, checked via `+dbug`. The player who should have received the concluding move still has the game stored in their active games.
>
> Fastest way to draw a game [here](https://www.chess.com/forum/view/fun-with-chess/fastest-stalemate).

Copied from [this comment](https://github.com/thecommons-urbit/chess/pull/148#issuecomment-1714716229).

ashelkovnykov:
Resolves #155

### DIFF
--- a/src/backend/chess/app/chess.hoon
+++ b/src/backend/chess/app/chess.hoon
@@ -480,7 +480,7 @@
             %=  this
               potential-states  (~(put by potential-states) game-id.action (snoc future-states u.move-result))
             ==
-          ::  did we win?
+          ::  is the game over?
           ?~  result.game.new-game-state
             ::  special draw available?
             ?:  ?&  special-draw-available.new-game-state
@@ -504,7 +504,7 @@
                     %chess-agent-action
                     !>([%receive-move game-id.action move.action])
             ==  ==
-          ::  tell opponent we won
+          ::  tell opponent checkmate or stalemate
           :~  :*  %pass
                   /poke/game/(scot %da game-id.action)/ended/[(need result.game.new-game-state)]
                   %agent
@@ -869,6 +869,11 @@
             ?:  =(result.action %'½–½')
               ::  is a draw now available?
               ?:  special-draw-available.result-game-state
+                (output-quip game.result-game-state(result `result.action) (bind move-result tail))
+              ::  does the move result in a draw?
+              ?:  ?&  ?=(^ result.game.new.u.move-result)
+                      ?=(%'½–½' u.result.game.new.u.move-result)
+                  ==
                 (output-quip game.result-game-state(result `result.action) (bind move-result tail))
               =/  san  (~(algebraicize with-position position.u.game-state) u.move.action)
               %+  poke-nack  this


src/backend/chess/app/chess.hoon
```diff
@@ -870,6 +870,11 @@
               ::  is a draw now available?
               ?:  special-draw-available.result-game-state
                 (output-quip game.result-game-state(result `result.action) (bind move-result tail))
+              ::  does the move result in a draw?
+              ?:  ?&  ?=(^ result.game.new.u.move-result)
+                      ?=(%'½–½' u.result.game.new.u.move-result)
+                  ==
```
bonbud-macryg, line 876: Why not something like this?

```hoon
::  does the move result in a draw?
?:  ?&  ?=([~ ?(%'1-0' %'0-1' %'½–½')] result.game.new.u.move-result)
        ?=(%'½–½' u.result.game.new.u.move-result)
    ==
```

bonbud-macryg (approved):
LGTM, one question aside. Behaves correctly on the frontend and the game is stored in both players' archives.

src/backend/chess/app/chess.hoon
ashelkovnykov: That should work also, but it's more verbose and we would need to change this line if we ever change how results are represented. If we don't, it would silently fail. So now we would need to remember that there is an unwritten relationship for the `result` field of the `chess-game` type in `sur` and `app`.

What we're trying to check is that:
1. the result after the move is a non-null `unit`
2. the value of the `unit` is a draw

We just need the compiler to narrow down the type of `result.game.new.u.move-result` to be "non-null `unit`", and it's become fairly idiomatic in Hoon to do it as above. Another way would be to use `?~`, but then we'd need to copy the nack code.

bonbud-macryg: That makes sense. Is `^` checking for *non-null*, though?

```
> (some ~)
[~ u=~]
> (some 1)
[~ u=1]
> ?=(^ (some ~))
%.y
> ?=(^ (some 1))
%.y
```

This would:

```hoon
::  does the move result in a draw?
?:  ?&  !=((some ~) result.game.new.u.move-result)
        ?=(%'½–½' u.result.game.new.u.move-result)
    ==
```

`!=([~ ~] result.game.new.u.move-result)` would also work, but you can argue as to which is clearer.

```
> !=((some ~) (some ~))
%.n
> !=((some ~) (some 1))
%.y
> !=([~ ~] (some ~))
%.n
> !=([~ ~] (some 1))
%.y
```

rovmug-ticfyn: I don’t think a lack of `result` is ever represented as `[~ ~]` though, just `~` (assuming fair play).

ashelkovnykov: @rovmug-ticfyn Correct

@bonbud-macryg I think there's a misunderstanding here. What we're trying to check is, "is there a result from applying the move, and is that result a draw?". The result has type `unit`. A `unit` can be either `~` or `[~ something]`. The first check is, "is the result after the move a cell?". If so, that rules out the case that result is `~`, since that's an atom. Therefore, we know definitively based on the intersection of the types `unit` and `cell` that result has the form `[~ something]`. The second check is then, "is the value of the result `unit` `%'½–½'`?".

bonbud-macryg: *Ohhh* right, for some reason I had it in my head that a "null unit" was `[~ ~]` and a non-null unit was `[~ u=something]`, but a "null unit" is just `~`. I checked [the docs](https://developers.urbit.org/reference/hoon/stdlib/1c#unit) and `unit` is just a gate spitting out `~` or `[~ u=item]` depending on if `item` is an atom or a cell, so of course checking if it's a cell works.

PR #162, opened 2023-11-02
Add new assets for boards and pieces

bonbud-macryg:
No integration, just uploading the files.


bonbud-macryg:
Clarification: there are boards here (`default.svg`, etc.) we intend to have a negative space effect as seen in mockups. These are using explicit "ivory"-coloured squares rather than using the transparent .svg to show the ivory background below in anticipation of an eventual dark mode.

For the dark mode the background would become the same colour as the black squares on `default.svg`; I think it would make sense to let the light mode's negative space effect play in reverse for that board and let the other boards stand out against the background. Transparent boards would create the effect in dark mode of having a red and black board, a green and black board, a black and black board in `default.svg`, etc.

ashelkovnykov (approved):

PR #163, opened 2023-12-31
marks: add missing .md mark

ashelkovnykov:
Fixes error on fresh install when it encounters `README.md`


ashelkovnykov:
The error was actually related to old, leftover files from a much earlier version of `%chess` in my desk.

PR #166, opened 2024-04-13
Add Vite

bonbud-macryg:
Vite automatically adds code changes to a server running the UI, and in this case proxies requests to that server to a running Urbit ship. This obviates the need to build and glob every single UI change through the globulator, and should be considered necessary for further frontend development on this app.

To make the app compatible with Vite this PR also improves the way `urbit-ob` is used, just importing its `isValidPatp` function, and adds a type declaration for that function to satisfy Typescript.

It only became apparent I'd have to fix `urbit-ob` when I clicked the Challenges tab and the screen went blank, so it's worth testing thoroughly through `npm run dev` to verify that nothing else will be broken for devs testing the app in Vite.


ashelkovnykov (approved):
LGTM

bonbud-macryg:
On reflection, clicking the Challenges tab was only broken because of the use of a `require()` to import the `urbit-ob` package rather than an `import`. This was the only `require()` in the codebase.

Given that, I think this is safe to merge without further testing; in any case an undiscovered issue with running the app in Vite would only affect devs testing the app in Vite.

